PR #1, opened 2021-11-16
Add initial charm structure

marceloneppel:
This PR introduces the following changes:

- README.md: Added repository description and usage instructions.
- Added the initial charm structure created by the `charmcraft init` call (with an initial PostgreSQL instance running - without replication - just to initialize the charm structure).

PR #2, opened 2021-11-23
Template structure

marceloneppel:
This PR introduces the following changes:

* Improved testing using layout from the [operator template](https://github.com/canonical/operator-template) and adding new unit and integration tests.
* Generated postgres user password instead of user provided password.
* Subscription to on pebble ready event in order to correctly fire the lifecycle events.
* Add storage.
* Moved testing and usage instructions to CONTRIBUTING.md.

Some items were discussed in #1.


.flake8
```diff
@@ -0,0 +1,9 @@
+[flake8]
+max-line-length = 99
+select: E,W,F,C,N
+exclude:
+  venv
+  .git
+  build
+  dist
+  *.egg_info
```
jnsgruk, line 9: This can be dropped and placed in the `pyproject.toml`

README.md
```diff
@@ -1 +1,23 @@
-Coming soon
+# PostgreSQL Kubernetes Operator
+
+## Overview
+
+This repository hosts the PostgreSQL Kubernetes Operator.
+
```
jnsgruk, line 6:
```suggestion
```

The overview section doesn't seem to add much value here :)

tox.ini
```diff
@@ -0,0 +1,79 @@
...
+  PY_COLORS=1
+passenv =
+  PYTHONPATH
+  HOME
+  PATH
+  CHARM_BUILD_DIR
+  MODEL_SETTINGS
+  HTTP_PROXY
+  HTTPS_PROXY
+  NO_PROXY
```
jnsgruk, line 28:
```suggestion
  PYTHONPATH
  HOME
  CHARM_BUILD_DIR
  MODEL_SETTINGS
```

These are passed through automatically I believe

jnsgruk (approved):
Looking good, some minor nits.

.flake8
marceloneppel: Thanks Jon. My mistake (I forgot to delete this file). Now it's fixed (file deleted and content present in `pyproject.toml`).

README.md
marceloneppel: I agree. I removed it as we already have a better overview in the description section.

src/charm.py
```diff
@@ -0,0 +1,103 @@
...
+        """Handle the config-changed event."""
+        # TODO: placeholder method to implement logic specific to configuration change.
+        pass
+
+    def _on_postgresql_pebble_ready(self, _) -> None:
+        """Event handler for on PebbleReadyEvent."""
+        # TODO: move this code to an "_update_layer" method in order to also utilize it in
+        # config-changed hook.
+        # Get the postgresql container so we can configure/manipulate it.
+        container = self.unit.get_container(self._postgresql_service)
```
DnPlas, line 49: I'm unsure about this one. If you deploy the application with a different name, this might not work. Something more generic could be a better option:

```python
self.unit.get_container(self.model.app.name)
```

It can also help differentiate between the name of the service and the container.

tox.ini
marceloneppel: You're right, Jon. These variables are passed by default in [Tox](https://github.com/tox-dev/tox/blob/4cf816c97de18dca9d99dc86fec341c1d2c4a20a/src/tox/config/__init__.py#L769).

src/charm.py
```diff
@@ -0,0 +1,103 @@
...
+        # Create a new config layer.
+        layer = self._postgresql_layer()
+
+        if container.can_connect():
+            # Get the current config.
+            services = container.get_plan().services
+            # Check if there are any changes to services.
+            if services != layer.services:
```
DnPlas, line 57: IMHO, we can improve readability and style with something like:

```python
#  These names make it more clear to understand what is what
new_layer = self._postgresql_layer().services
if container.can_connect():
    current_layer = container.get_plan().services

# Now the statement compares two of the same kind
if current_layer != new_layer:
    ...
```

src/charm.py
```diff
@@ -0,0 +1,103 @@
...
+                        # We need to set either POSTGRES_HOST_AUTH_METHOD or POSTGRES_PASSWORD
+                        # in order to initialize the database.
+                        "POSTGRES_PASSWORD": self._stored.postgres_password,
+                    },
+                }
+            },
+        }
+        return Layer(layer_config)
+
+    def _new_password(self):
```
DnPlas, line 91: Minor suggestion since you are using type annotations.

```suggestion
    def _new_password(self) -> str:
       ...
```

DnPlas (commented):
Just few comments.

src/charm.py
marceloneppel: Thanks for suggesting this again, Daniela. I'll try to remember in the next changes. Committed the change.

src/charm.py
marceloneppel: Good suggestion. Thanks, Daniela. Testing, seems like that using the app name also may cause the charm to not find the container if we use a different application name.

Maybe the best approach should be using the `workload` property from the `event` argument. What do you think @jnsgruk? (I saw `event.workload` in the Kubernetes dashboard charm) Maybe it can be something like:

```python
def _on_postgresql_pebble_ready(self, event: WorkloadEvent) -> None:
    """Event handler for on PebbleReadyEvent."""
    container = event.workload
```

src/charm.py
```diff
@@ -0,0 +1,103 @@
...
+        self._postgresql_service = "postgresql"
+
+        self.framework.observe(self.on.install, self._on_install)
+        self.framework.observe(self.on.config_changed, self._on_config_changed)
+        self.framework.observe(self.on.postgresql_pebble_ready, self._on_postgresql_pebble_ready)
+
+    def _on_install(self, _):
+        """Event handler for InstallEvent."""
+        # TODO: change to peer/leader data bag when relations are implemented.
+        self._stored.postgres_password = self._new_password()
```
MiaAltieri, line 37: Correct me if I am wrong, but I think this can be done without relations being implemented (this is because peers refer to units from the same application as according to the [Juju SDK](https://juju.is/docs/sdk/relations#heading--peer-relations), I believe `self.peers` should still be accessible and can be used to save passwords by doing something like:

```
        root_password = None

        data = self.peers.data[self.peers.app]
        root_password = data.get('root_password', None)

        if root_password is None:
            root_password = self._new_password()
            <set peer password>
```

metadata.yaml
```diff
@@ -0,0 +1,27 @@
...
+resources:
+  postgresql-image:
+    type: oci-image
+    description: OCI image for PostgreSQL (ubuntu/postgres)
+    upstream-source: ubuntu/postgres:12-20.04_beta
+
+storage:
+  pgdata:
+    type: filesystem
+    location: /var/lib/postgresql/data/pgdata
```
MiaAltieri, line 27: Might want to add `peer` info here
```
peers:
  postgresql:
    interface: postgresql
```

src/charm.py
jnsgruk: Yep, you can definitely do that inside the `pebble-ready` events -- obviously doesn't work in any other kind of event handler but an easy way to make sure you're getting the right container here :)

src/charm.py
jnsgruk: This is correct - so long as you have a `peers` section in metadata :)

One thing worth noting, is that you'll want to guard writes to the `self.peers.data[self.app]` attribute with `if self.unit.is_leader()`. Similarly to provides/requires relation, units can write to their own data bag, but only leaders can write to the *application* data bag :)

metadata.yaml
jnsgruk: Agreed - might be worth naming it `postgresql-replicas` or similar, though

src/charm.py
marceloneppel: Great. I updated the code to use `event.workload`. Thanks for the note about the use only on `pebble-ready` event.

src/charm.py
marceloneppel: Thanks, Daniela! Yes, I agree that it gets more readable if we do that changes. I updated the code and changed something in your suggestion in order to make it clear for others that the variables represent layers/plan instead of services (and be able use `new_layer` in the `add_layer` call):

```python
new_layer = self._postgresql_layer()
if container.can_connect():
    current_layer = container.get_plan()
    if current_layer.services != new_layer.services:
        container.add_layer(self._postgresql_service, new_layer, combine=True)
```

src/charm.py
marceloneppel: Thanks for the great contribution Mia and Jon. I used the `_on_leader_elected` event in order to enable just the leader to write the data. The changes are in the last commit.

metadata.yaml
marceloneppel: Thanks Mia and Jon. I created the peer relation and submitted in the last commit.

jnsgruk (approved):
Good start, let's get this merged and move on, I think :)

actions.yaml
```diff
@@ -0,0 +1,5 @@
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+get-postgres-password:
```
jnsgruk, line 4: One minor change

```suggestion
get_postgres_password:
```

marceloneppel: This will cause an `bad action name` error on deploy (we need to follow the action name rule from https://github.com/juju/charm/blob/a407f299325d3fcfdd249a9c104cb87184d61064/actions.go#L20).

PR #4, opened 2022-02-11
DPE-8 postgresql k8s replication with automatic failover

marceloneppel:
This PR enables the replication of the data among PostgreSQL units and also automatic failover, both features using [Patroni](https://github.com/zalando/patroni).

The integration tests focus on checking what is needed (configuration files and labels) to make Patroni manage correctly the replication and the automatic failover.

There is also a workaround in the update status hook to restart Patroni + PostgreSQL during an automatic failover. It’s needed because there is an issue about [zombie processes](https://github.com/canonical/pebble/issues/6).

Even separating some manual actions (manual failover and switchover) from this PR it got a little bigger. I’ll work to have smaller PRs in the future.

One last detail: the integration tests should fail as the docker image is not published yet, but it can be executed locally using the instructions from CONTRIBUTING.md.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -161,10 +161,12 @@ class PostgresqlOperatorCharm(CharmBase):
                 # Restart the stuck replica.
                 self._restart_postgresql_service()
             elif state == "stopping":
-                # Restart the stuck previous leader (forcing the immediate  failover).
+                # Change needed to force an immediate failover.
                 self._patroni.change_master_start_timeout(0)
+                # Restart the stuck previous leader (will trigger an immediate failover).
                 self._restart_postgresql_service()
-                self._patroni.change_master_start_timeout(300)
+                # Revert configuration to default.
+                self._patroni.change_master_start_timeout(None)
         except RetryError as e:
             logger.error("failed to check PostgreSQL state")
             self.unit.status = BlockedStatus(f"failed to check PostgreSQL state with error {e}")
--- a/src/charm.py
+++ b/src/charm.py
@@ -27,7 +27,6 @@ from patroni import Patroni
 logger = logging.getLogger(__name__)
 
 PEER = "postgresql-replicas"
-STORAGE_PATH = "/var/lib/postgresql/data"
 
 
 class PostgresqlOperatorCharm(CharmBase):
@@ -52,7 +51,8 @@ class PostgresqlOperatorCharm(CharmBase):
         )
         self.framework.observe(self.on.get_primary_action, self._on_get_primary)
         self.framework.observe(self.on.update_status, self._on_update_status)
-        self._patroni = Patroni(self._pod_ip)
+        self._storage_path = self.meta.storages["pgdata"].location
+        self._patroni = Patroni(self._pod_ip, self._storage_path)
 
     def _on_install(self, _) -> None:
         """Event handler for InstallEvent."""
@@ -96,9 +96,9 @@ class PostgresqlOperatorCharm(CharmBase):
                 # Changes were made, add the new layer.
                 container.add_layer(self._postgresql_service, new_layer, combine=True)
                 logging.info("Added updated layer 'postgresql' to Pebble plan")
-                self._patroni.render_patroni_yml_file()
                 # TODO: move this file generation to on config changed hook
                 # when adding configs to this charm.
+                self._patroni.render_patroni_yml_file()
                 # Restart it and report a new status to Juju.
                 container.restart(self._postgresql_service)
                 logging.info("Restarted postgresql service")
@@ -198,7 +198,7 @@ class PostgresqlOperatorCharm(CharmBase):
                 self._postgresql_service: {
                     "override": "replace",
                     "summary": "entrypoint of the postgresql + patroni image",
-                    "command": f"/usr/bin/python3 /usr/local/bin/patroni {STORAGE_PATH}/patroni.yml",
+                    "command": f"/usr/bin/python3 /usr/local/bin/patroni {self._storage_path}/patroni.yml",
                     "startup": "enabled",
                     "user": "postgres",
                     "group": "postgres",
--- a/src/charm.py
+++ b/src/charm.py
@@ -7,7 +7,6 @@
 import logging
 import secrets
 import string
-from time import sleep
 
 from lightkube import ApiError, Client, codecs
 from lightkube.resources.core_v1 import Pod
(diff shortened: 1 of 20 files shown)


src/charm.py
```diff
@@ -7,14 +7,29 @@
...
+)
 from ops.pebble import Layer
+from tenacity import RetryError
+
+from patroni import Patroni
 
 logger = logging.getLogger(__name__)
 
+PEER = "postgresql-replicas"
+STORAGE_PATH = "/var/lib/postgresql/data"
```
paulomach, line 31: @marceloneppel 
noob question: since the storage_path is being defined in the metadata, could it be retrieved in some different manner to avoid duplication?

src/resources.yaml
```diff
@@ -0,0 +1,85 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ app_name }}
```
paulomach, line 4: is the {{ app_name }} replaced on the go? do we use jinja or something like it?

src/charm.py
```diff
@@ -46,12 +72,16 @@ def _on_leader_elected(self, _) -> None:
         """Handle the leader-elected event."""
         data = self._peers.data[self.app]
```
paulomach, line 73: is this data shared across all peers?

src/charm.py
```diff
@@ -67,6 +97,9 @@ def _on_postgresql_pebble_ready(self, event: WorkloadEvent) -> None:
                 # Changes were made, add the new layer.
                 container.add_layer(self._postgresql_service, new_layer, combine=True)
                 logging.info("Added updated layer 'postgresql' to Pebble plan")
+                self._patroni.render_patroni_yml_file()
+                # TODO: move this file generation to on config changed hook
+                # when adding configs to this charm.
                 # Restart it and report a new status to Juju.
                 container.restart(self._postgresql_service)
```
paulomach, line 104: Why the restart is required?

src/charm.py
```diff
@@ -119,13 +235,29 @@ def _peers(self) -> Relation:
...
+        return data.get("replication-password", None)
+
+    def _unit_name_to_pod_name(self, unit_name: str) -> str:
+        """Converts unit name to pod name.
+
+        Args:
+            unit_name: name in "postgresql-k8s/0" format.
+
+        Returns:
+            pod name in "psotgresql-k8s-0" format.
```
paulomach, line 257: @marceloneppel , minor typo here. s/psotgresql/postgresql/

src/patroni.py
```diff
@@ -0,0 +1,99 @@
...
+import pwd
+
+import requests
+from jinja2 import Template
+from tenacity import retry, stop_after_attempt, wait_exponential
+
+logger = logging.getLogger(__name__)
+
+
+STORAGE_PATH = "/var/lib/postgresql/data"
```
paulomach, line 18: Same as before. Would be nice to have a single place for this constant

templates/patroni.yml.j2
```diff
@@ -0,0 +1,27 @@
...
+log:
+  dir: /var/log/postgresql
+restapi:
+  connect_address: '{{ pod_ip }}:8008'
+  listen: 0.0.0.0:8008
+pod_ip: '{{ pod_ip }}'
+postgresql:
+  connect_address: '{{ pod_ip }}:5432'
+  custom_conf: /var/lib/postgresql/data/postgresql-k8s-operator.conf
+  data_dir: /var/lib/postgresql/data/pgdata
```
paulomach, line 24: @marceloneppel is the data_dir equivalent to the storage_path?

src/resources.yaml
paulomach: nevermind, found it

src/charm.py
jnsgruk: Good call - you might be able to use the `CharmMeta` class in `ops` for this

src/charm.py
jnsgruk: Yep!

src/charm.py
```diff
@@ -75,26 +108,109 @@ def _on_postgresql_pebble_ready(self, event: WorkloadEvent) -> None:
...
 
+    def _on_get_primary(self, event: ActionEvent) -> None:
+        """Get primary instance."""
+        try:
+            primary = self._patroni.get_primary(unit_name_pattern=True)
+            event.set_results({"primary": primary})
+        except RetryError as e:
+            logger.error(f"failed to get primary with error {e}")
+
+    def _on_update_status(self, _) -> None:
```
jnsgruk, line 155: Might it be an idea to set the `ActiveStatus` message in here if the current unit is the primary? Something like:

```python
if self._patroni.get_primary(unit_name_pattern=True) == self.unit.name:
	self.unit.status = ActiveStatus("Primary")
```

src/charm.py
```diff
@@ -75,26 +108,109 @@ def _on_postgresql_pebble_ready(self, event: WorkloadEvent) -> None:
...
+
+        Args:
+            wait: whether or not to wait 30 seconds between stop and start.
+        """
+        self.unit.status = MaintenanceStatus(f"restarting {self._postgresql_service} service")
+        container = self.unit.get_container("postgresql")
+        container.stop(self._postgresql_service)
+        # Sleep needed to make the leader release the lock in a failover/switchover.
+        if wait:
+            sleep(30)
```
jnsgruk, line 182: Hmm - I don't like this at all - sleeping in an hook blocks all other hook invocations/model operations for this charm. You could use `tenacity` here to just block on a specific condition if it's a must, but arbitrary sleeps for 30s is not a good look.

src/charm.py
```diff
@@ -119,13 +235,29 @@ def _peers(self) -> Relation:
...
         """
-        return self.model.get_relation("postgresql-replicas")
+        return self.model.get_relation(PEER)
 
     def _get_postgres_password(self) -> str:
         """Get postgres user password."""
         data = self._peers.data[self.app]
         return data.get("postgres-password", None)
 
+    def _get_replication_password(self) -> str:
```
jnsgruk, line 245:
```suggestion
    @property
    def _replication_password(self) -> str:
```

jnsgruk (changes requested):
Overall this is looking really sharp! Let's have a bit of a think about that sleep though...

src/charm.py
marceloneppel: Thanks for catching it @paulomach! I'll fix it.

src/charm.py
marceloneppel: Thanks for the feedback Jon! I definitely agree. I'll improve that with your suggestion to avoid blocking the other hooks.

src/charm.py
marceloneppel: The restart is required at this point to really start patroni/postgresql (their service was not yet present in the container until we add the new layer on L98).

paulomach: Got it. Thanks for the explainer.

src/patroni.py
marceloneppel: I agree. Thanks Paulo!

templates/patroni.yml.j2
marceloneppel: `data_dir` is a folder inside `storage_path`. Your comment made me think about maybe I can use the STORAGE_PATH parameter here to not duplicate the path again. Thanks for the comment.

jnsgruk (approved):

paulomach (approved):
LGTM, but there's an error in the integration tests (don't know if it was occurring before though).

marceloneppel:
> LGTM, but there's an error in the integration tests (don't know if it was occurring before though).

Thanks @paulomach! Yes, these tests are failing because the docker image is not published anywhere (to pull it while testing in the CI).

PR #6, opened 2022-04-05
DPE-124 Additional Integration Testing 

WRFitch:
## Issue
This PR adds some integration tests that verify that data persists in the postgres charm after graceful restarts and catastrophic failures.

## Context
- Integration tests are failing on GitHub because the test runner doesn't have access to the patroni container.

## Release Notes
- Extra documentation for setup steps
- some minor cleanup for other tests, including `get_postgres_password` and `db_connect` methods
- Some extra `wait_for_idle`s in other tests to guarantee stability
- `test_persist_data_through_graceful_restart` test, which tests that data written to the primary persists across the other nodes after scaling juju down to 0 and back up to 3 nodes
- `test_persist_data_through_failure` test, which tests that data written to the primary persists across the other nodes after killing the primary using microk8s
- Early return in `_on_update_status` hook handler when postgres is inaccessible
- Added pytest logging to show how long each integration test runs for

### DIFF
--- a/tests/integration/test_charm.py
+++ b/tests/integration/test_charm.py
@@ -174,7 +174,12 @@ async def test_persist_data_through_failure(ops_test: OpsTest):
     # Wait for juju to notice one of the pods is gone and fix it
     logger.info("wait for juju to reset postgres container")
     await ops_test.model.wait_for_idle(
-        apps=[APP_NAME], status="active", timeout=1000, wait_for_exact_units=3, check_freq=2, idle_period=45
+        apps=[APP_NAME],
+        status="active",
+        timeout=1000,
+        wait_for_exact_units=3,
+        check_freq=2,
+        idle_period=45,
     )
     logger.info("juju has reset postgres container")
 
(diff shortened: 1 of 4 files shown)


tests/integration/test_charm.py
```diff
@@ -177,3 +261,27 @@ async def pull_content_from_unit_file(unit, path: str) -> str:
     """
     action = await unit.run(f"cat {path}")
     return action.results.get("Stdout", None)
+
+
+async def wait_for_status(ops_test, status: str, wait: float):
+    status = False
```
shayancanonical, line 267: why overwrite the input value `status` to `False`?

tests/integration/test_charm.py
```diff
@@ -177,3 +261,27 @@ async def pull_content_from_unit_file(unit, path: str) -> str:
     """
     action = await unit.run(f"cat {path}")
     return action.results.get("Stdout", None)
+
+
+async def wait_for_status(ops_test, status: str, wait: float):
```
shayancanonical, line 266: is `wait_for_status` being used anywhere? if not, should we delete it?

WRFitch: Nope, this is totally unrelated and shouldn't be here.

marceloneppel (approved):
Great work Will! LGTM!

PR #9, opened 2022-05-05
Add scale-up/down logic

marceloneppel:
# Issue
In short this PR partially address JIRA ticket [DPE-161](https://warthogs.atlassian.net/browse/DPE-161) and [DPE-162](https://warthogs.atlassian.net/browse/DPE-162).

In long:
* To avoid overload of the network during copy of data between the replicas that are being added to the cluster during a scale up or downtime in the case of a scale down, the charm needs to add and remove one member at a time from the cluster.
* We can't run a manual switchover to switch the leader to a health unit during the departing of an unit because of a limitation in the moment the charm receives the event (a failover automatically happens after Patroni gets the information that the unit has gone).

# Solution
* Wait for all units to be ready (with `running` state in the Patroni REST API) before adding one more unit to the cluster.
* Only remove the unit from the list of current units when the charm receives a relation departed event, as the unit isn't accessible anymore (we receive the event after the unit has completely gone).

# Context
* Also changed the charm to use k8s host names instead of IP addresses.
* Pod IP is still needed only during the replica bootstrap process (it's replaced by the host names later).
* When a unit is departing as a result of a scale down process and this unit is the leader Patroni itself triggers a failover automatically. The best approach would be to make the charm call it or a switchover before the unit is removed, but, as commented above, currently there is a limitation related to the moment the charm receives the relation departed event on k8s.
* Moved `use endpoints` variable to environment variables as it was not working on `patroni.yml`.
* The list of current host names (endpoints) is tracked on application relation data to avoid reaching Patroni REST API every time this information is needed.
* Replicas are started pointing to the IP of the primary (to avoid they thinking they need to bootstrap a new separated cluster) and they get some additional pod labels later (which are needed to start the bootstrap of the replica in the cluster and make the replication works later), when it's ok to add them to the cluster (one at a time).
* During scale up, the status of the new member is not checked, only the status of the current members (they must be running/ready). This is done this way because Patroni cannot be started without pointing a leader, otherwise it'll create a new cluster in the replica, and when started in this way it has some changed of the state, like stopped and creating replica, so for now it's better to not check these status, as they may conflict with the status that we have when it really becomes part of the existing cluster.

# Testing
* Unit and integrations tests were updated.
* The current integration tests assert that the data is still being replicated after scaling down and up again.
* The test which asserts the database is up was changed to check Patroni REST API (which ensures PostgreSQL is up and running).
* The test which asserts the correct settings in the database was changed to check the information in the database itself instead of checking only configuration files (so we can get the currently applied config).
* Some more units tests need to be implemented in a subsequent PR to improve the coverage.

# Release Notes
* Change charm to use host names instead of IP addresses.
* Implement the correct scaling up and down mechanism.


WRFitch (changes requested):

src/charm.py
```diff
@@ -259,10 +510,23 @@ def _unit_name_to_pod_name(self, unit_name: str) -> str:
...
-            pod name in "psotgresql-k8s-0" format.
+            pod name in "postgresql-k8s-0" format.
         """
         return unit_name.replace("/", "-")
 
+    def force_primary_change(self):
+        """Force primary changes immediately.
+
+        This function is needed to handle cases related to
+        https://github.com/canonical/pebble/issues/6.
```
MiaAltieri, line 521: period leads to weird linking
```suggestion
        https://github.com/canonical/pebble/issues/6 .
```

src/charm.py
```diff
@@ -259,10 +510,23 @@ def _unit_name_to_pod_name(self, unit_name: str) -> str:
...
         Returns:
-            pod name in "psotgresql-k8s-0" format.
+            pod name in "postgresql-k8s-0" format.
         """
         return unit_name.replace("/", "-")
 
+    def force_primary_change(self):
+        """Force primary changes immediately.
+
+        This function is needed to handle cases related to
```
MiaAltieri, line 520: provide a 1 sentence summary as to why (and keep the link)

src/patroni.py
```diff
@@ -61,9 +71,32 @@ def get_postgresql_state(self) -> str:
...
+    def is_all_members_ready(self) -> bool:
+        """Check if all members are correctly running Patroni and PostgreSQL."""
+        # Request info from cluster endpoint
+        # (which returns all members of the cluster and their states).
+        r = requests.get(f"http://{self._endpoint}:8008/cluster")
+        return all(member["state"] == "running" for member in r.json()["members"])
+
+    @property
+    @retry(stop=stop_after_attempt(3), wait=wait_exponential(multiplier=1, min=2, max=10))
+    def member_started(self) -> bool:
```
MiaAltieri, line 95: see comment for restructure of this function

src/patroni.py
```diff
@@ -61,9 +71,32 @@ def get_postgresql_state(self) -> str:
...
+        r = requests.get(f"http://{self._endpoint}:8008/cluster")
+        return set([member["name"] for member in r.json()["members"]])
+
+    @retry(stop=stop_after_attempt(3), wait=wait_exponential(multiplier=1, min=2, max=10))
+    def is_all_members_ready(self) -> bool:
+        """Check if all members are correctly running Patroni and PostgreSQL."""
+        # Request info from cluster endpoint
+        # (which returns all members of the cluster and their states).
+        r = requests.get(f"http://{self._endpoint}:8008/cluster")
+        return all(member["state"] == "running" for member in r.json()["members"])
```
MiaAltieri, line 91: might want to restructure in a similar pattern to the suggestion I have for `member_started`

src/charm.py
```diff
@@ -68,7 +123,75 @@ def _on_config_changed(self, _):
...
+        """Add member to the cluster if all members are already up and running.
+
+        Raises:
+            NotReadyError if either the new member or the current members are not ready.
+        """
+        hostname = self._get_hostname_from_unit(member)
+
+        if not self._patroni.is_all_members_ready():
+            logger.info("not all members are ready")
+            raise NotReadyError("not all members are ready")
```
MiaAltieri, line 161: Restructure this in a similar way to my suggestion for `member_started `

src/patroni.py
```diff
@@ -106,3 +145,22 @@ def render_postgresql_conf_file(self) -> None:
...
+        self.render_patroni_yml_file()
+
+        try:
+            if self.member_started:
+                # Make Patroni use the updated configuration.
+                self._reload_patroni_configuration()
+        except RetryError:
+            # Ignore retry errors that happen when the member has not started yet.
+            # The configuration will be loaded correctly when Patroni starts.
+            pass
```
MiaAltieri, line 161: some functions have `try` `except` on this function for `RetryError`, so you might either want to raise this or remove those `except`s

tests/integration/test_charm.py
```diff
@@ -244,18 +248,18 @@ async def get_postgres_password(ops_test: OpsTest):
     return result.results["postgres-password"]
 
 
-async def pull_content_from_unit_file(unit, path: str) -> str:
-    """Pull the content of a file from one unit.
+async def get_unit_address(ops_test: OpsTest, unit_name: str):
```
MiaAltieri, line 251: put all non tests in a file `tests/integration/helpers.py` (can be done in subsequent PR but don't forget about it)

tests/integration/test_charm.py
```diff
@@ -4,10 +4,11 @@
 
 import logging
 import os
+from typing import List
```
MiaAltieri, line 7: add tests for partial scale up and partial scale down (you have full scale down which is great! But a partial would also be nice)

tests/unit/test_charm.py
```diff
@@ -31,21 +31,18 @@ def setUp(self):
             "app_name": self.harness.model.app.name,
         }
 
+    @patch_network_get(private_address="1.1.1.1")
```
MiaAltieri, line 34: I'll review the unit tests in a separate review once the issues covered in `patroni.py` and `charm.py` are covered

tests/unit/test_patroni.py
```diff
@@ -5,36 +5,40 @@
 import unittest
 from unittest.mock import mock_open, patch
 
+from jinja2 import Template
```
MiaAltieri, line 8: I'll review the unit tests in a separate review once the issues covered in patroni.py and charm.py are covered

src/charm.py
```diff
@@ -45,21 +53,68 @@ def __init__(self, *args):
...
+        old = self._endpoints
+        current = [self._get_hostname_from_unit(member) for member in self._hosts]
+        endpoints_to_remove = list(set(old) - set(current))
+        return endpoints_to_remove
+
+    def _on_peer_relation_departed(self, _):
+        """The leader removes the departing units from the list of cluster members."""
+        if not self.unit.is_leader():
+            return
+
```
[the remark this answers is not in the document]
paulomach: I say it's good practice to keep idempotency

src/charm.py
```diff
@@ -89,30 +229,52 @@ def _on_postgresql_pebble_ready(self, event: WorkloadEvent) -> None:
...
+        if not container.can_connect():
             self.unit.status = WaitingStatus("waiting for Pebble in workload container")
+            event.defer()
+            return
+
+        # Defer the initialization of the workload in the replicas
+        # if the cluster hasn't been bootstrap on the primary yet.
+        if not self.unit.is_leader() and "cluster_initialised" not in self._peers.data[self.app]:
+            event.defer()
+            return
```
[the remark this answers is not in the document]
paulomach: +1 here. If the member are added on the `relation_changed`, why delay initialisation?

paulomach (commented):
Mia's comments are pointing the same things I've found.

src/charm.py
```diff
@@ -45,21 +53,68 @@ def __init__(self, *args):
...
+            try:
+                # Update the cluster members in this unit (updating patroni configuration).
+                self._patroni.update_cluster_members()
+                # Add the labels needed for replication in this pod.
+                self._patch_pod_labels(self._unit)
+                # Validate the status of the member before setting an ActiveStatus.
+                try:
+                    if not self._patroni.member_started:
+                        raise NotReadyError
+                except (NotReadyError, RetryError):
```
shayancanonical, line 108: would it be possible to avoid having nested try/except blocks? these are usually really hard to trace and debug if need be in the future

tests/integration/test_charm.py
marceloneppel: I agree Mia. Implemented on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/3651c6d458a88df4287e8c0d389ce2bcac9fec76.

src/patroni.py
marceloneppel: The other functions doesn't have try/except on this function anymore.

src/charm.py
```diff
@@ -68,7 +123,75 @@ def _on_config_changed(self, _):
         # TODO: placeholder method to implement logic specific to configuration change.
         pass
 
-    def _on_leader_elected(self, _) -> None:
+    def _reconfigure(self, event) -> None:
```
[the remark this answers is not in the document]
marceloneppel: It seems better. Changed on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/ee0018899c32c6f9b60debcc5179ef39eb6ac02c.

src/charm.py
```diff
@@ -68,7 +123,75 @@ def _on_config_changed(self, _):
...
+        """Reconfigure cluster members."""
```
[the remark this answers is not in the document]
marceloneppel: Added on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/ee0018899c32c6f9b60debcc5179ef39eb6ac02c.

src/charm.py
```diff
@@ -45,21 +53,68 @@ def __init__(self, *args):
...
         self.framework.observe(self.on.upgrade_charm, self._on_upgrade_charm)
         self.framework.observe(
             self.on.get_postgres_password_action, self._on_get_postgres_password
         )
         self.framework.observe(self.on.get_primary_action, self._on_get_primary)
         self.framework.observe(self.on.update_status, self._on_update_status)
         self._storage_path = self.meta.storages["pgdata"].location
-        self._patroni = Patroni(self._pod_ip, self._storage_path)
+
+    @property
```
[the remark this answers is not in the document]
marceloneppel: Fixed on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/ee0018899c32c6f9b60debcc5179ef39eb6ac02c.

src/charm.py
```diff
@@ -45,21 +53,68 @@ def __init__(self, *args):
...
+    def _endpoints_to_remove(self):
```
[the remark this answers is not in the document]
marceloneppel: Fixed on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/ee0018899c32c6f9b60debcc5179ef39eb6ac02c.

src/charm.py
```diff
@@ -45,21 +53,68 @@ def __init__(self, *args):
...
+                try:
+                    if not self._patroni.member_started:
+                        raise NotReadyError
+                except (NotReadyError, RetryError):
+                    self.unit.status = WaitingStatus("awaiting for member to start")
+                    event.defer()
+                    return
+                self.unit.status = ActiveStatus()
+            except RetryError:
+                self.unit.status = BlockedStatus("failed to update cluster members on member")
```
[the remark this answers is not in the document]
marceloneppel: Great idea Mia! Thanks for sharing this different way of retry. Changed on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/ee0018899c32c6f9b60debcc5179ef39eb6ac02c.

src/charm.py
```diff
@@ -152,6 +321,51 @@ def _on_get_primary(self, event: ActionEvent) -> None:
         except RetryError as e:
             logger.error(f"failed to get primary with error {e}")
 
+    def _on_stop(self, _) -> None:
+        """Handle the stop event."""
```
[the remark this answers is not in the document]
marceloneppel: Moving this function to another PR as your comment about splitting in 2/3 PRs makes sense.

src/charm.py
```diff
@@ -152,6 +321,51 @@ def _on_get_primary(self, event: ActionEvent) -> None:
         except RetryError as e:
             logger.error(f"failed to get primary with error {e}")
 
+    def _on_stop(self, _) -> None:
```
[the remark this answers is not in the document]
marceloneppel: Moving this function to another PR as your comment about splitting in 2/3 PRs makes sense.

src/charm.py
```diff
@@ -161,13 +375,8 @@ def _on_update_status(self, _) -> None:
...
                 self._restart_postgresql_service()
-            elif state == "stopping":
-                # Change needed to force an immediate failover.
-                self._patroni.change_master_start_timeout(0)
-                # Restart the stuck previous leader (will trigger an immediate failover).
-                self._restart_postgresql_service()
-                # Revert configuration to default.
-                self._patroni.change_master_start_timeout(None)
+            elif state == "starting" or state == "stopping":
+                self.force_primary_change()
```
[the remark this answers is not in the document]
marceloneppel: This is needed because a bug on Pebble. It'll be removed along with other functions in the next PR as the current version of Juju solved the problem of zombie process from PostgreSQL not being cleaned by Pebble.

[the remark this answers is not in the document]
marceloneppel: This is needed because a bug on Pebble. It'll be removed along with other functions in the next PR as the current version of Juju solved the problem of zombie process from PostgreSQL not being cleaned by Pebble.

src/charm.py
marceloneppel: Fixed on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/ee0018899c32c6f9b60debcc5179ef39eb6ac02c.

src/patroni.py
marceloneppel: Changed on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/ee0018899c32c6f9b60debcc5179ef39eb6ac02c.

src/patroni.py
marceloneppel: Great. Changed on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/ee0018899c32c6f9b60debcc5179ef39eb6ac02c.

src/charm.py
marceloneppel: I kept it this way because the unit status is being changed to blocked in the function that calls this function.

tests/integration/test_charm.py
marceloneppel: Of course. The new functions were already created in the helpers and the remaining will be moved to there.

src/charm.py
marceloneppel: Added more info on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/ee0018899c32c6f9b60debcc5179ef39eb6ac02c, This should be removed in the next PR as the problem that it's related too was solved in the current Juju version.

src/charm.py
```diff
@@ -68,7 +123,75 @@ def _on_config_changed(self, _):
...
+            raise NotReadyError("not all members are ready")
+
+        # Add the member to the list that should be updated in each other member.
+        self._update_endpoints(endpoint_to_add=hostname)
+        # Add the required labels for replication to the member pod.
+        self._patch_pod_labels(member)
+
+        # Update the list of members in this unit (updating the Patroni configuration).
+        try:
+            self._patroni.update_cluster_members()
```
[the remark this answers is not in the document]
marceloneppel: Fixed on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/ee0018899c32c6f9b60debcc5179ef39eb6ac02c.

src/charm.py
```diff
@@ -45,21 +53,68 @@ def __init__(self, *args):
...
+        # one at a time.
+        if self.unit.is_leader():
+            self._reconfigure(event)
+            return
+
+        if self._endpoint in self._endpoints:
+            # Update the list of the cluster members in the replicas to make them know each other.
+            try:
+                # Update the cluster members in this unit (updating patroni configuration).
+                self._patroni.update_cluster_members()
```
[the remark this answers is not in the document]
marceloneppel: Fixed on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/ee0018899c32c6f9b60debcc5179ef39eb6ac02c. Thanks for catching it!

src/charm.py
```diff
@@ -45,21 +53,68 @@ def __init__(self, *args):
...
+        # before any other member joins the cluster.
+        if "cluster_initialised" not in self._peers.data[self.app]:
+            event.defer()
+            return
+
+        # If the leader is the one receiving the event, it adds the new members,
+        # one at a time.
+        if self.unit.is_leader():
+            self._reconfigure(event)
+            return
```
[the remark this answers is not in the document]
marceloneppel: You're right. I removed the return and also the duplicated code in the `add_cluster_member` function, that is executed here.

src/charm.py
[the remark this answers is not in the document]
marceloneppel: Thanks. It seems better to have this check. Implemented in https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/ee0018899c32c6f9b60debcc5179ef39eb6ac02c.

src/charm.py
```diff
@@ -45,21 +53,68 @@ def __init__(self, *args):
...
+
+        if self._endpoint in self._endpoints:
```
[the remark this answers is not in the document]
marceloneppel: We need the endpoints in the endpoints property because is this property that tells the members that were added to the cluster and need to be updated in the cluster members list of the patroni configuration.

src/charm.py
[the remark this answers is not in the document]
marceloneppel: I added a better comment about that on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/ee0018899c32c6f9b60debcc5179ef39eb6ac02c. It's needed because if multiple units start Patroni at the same time they will create multiple clusters (separated from each other). We need to start the primary first and link the other units to it (through the cluster members list) before starting.

src/charm.py
```diff
@@ -80,6 +203,23 @@ def _on_leader_elected(self, _) -> None:
         if replication_password is None:
             self._peers.data[self.app]["replication-password"] = self._new_password()
 
+        # Create resources and add labels needed for replication.
+        self._create_resources()
+
+        # Patch the pod labels of this unit to enable replication later.
+        self._patch_pod_labels(self.unit.name)
```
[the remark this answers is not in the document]
marceloneppel: If we keep it in on install it will add the replicas to the cluster too soon (in a unknown state that prevents us from checking later that all members are ready; we can create some logic to handle this unknown state in the `are_all_members_ready` function but it could generate some confusion for someone that is reading the code). So, it's better to add the pod labels later. I moved it to both pebble ready and add cluster member functions on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/ee0018899c32c6f9b60debcc5179ef39eb6ac02c, so it's done only once per unit.

tests/unit/test_charm.py
```diff
@@ -92,6 +93,8 @@ def test_on_postgresql_pebble_ready(self, _render_patroni_yml_file):
 
         # Get the current and the expected layer from the pebble plan and the _postgresql_layer
         # method, respectively.
+        # TODO: test also replicas.
```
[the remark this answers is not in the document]
marceloneppel: Added on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/c7043a0579e3540855a2c33ce5ec297c321b18b2.

src/charm.py
```diff
@@ -121,7 +283,8 @@ def _patch_pod_labels(self) -> None:
             }
             client.patch(
                 Pod,
-                name=self._unit_name_to_pod_name(self._unit),
+                # name=self._unit_name_to_pod_name(self._unit),
```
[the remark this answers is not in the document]
marceloneppel: No. Thanks for catching it. Removed on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/c7043a0579e3540855a2c33ce5ec297c321b18b2,

src/charm.py
```diff
@@ -89,30 +229,52 @@ def _on_postgresql_pebble_ready(self, event: WorkloadEvent) -> None:
...
-                # TODO: move this file generation to on config changed hook
-                # when adding configs to this charm.
-                self._patroni.render_patroni_yml_file()
-                # Restart it and report a new status to Juju.
-                container.restart(self._postgresql_service)
-                logging.info("Restarted postgresql service")
-            # All is well, set an ActiveStatus.
-            self.unit.status = ActiveStatus()
-        else:
+        # Defer the event if we pebble is not available yet.
```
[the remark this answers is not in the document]
marceloneppel: Fixed on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/c7043a0579e3540855a2c33ce5ec297c321b18b2.

src/charm.py
```diff
@@ -188,9 +397,50 @@ def _restart_postgresql_service(self) -> None:
...
+    def _endpoints(self) -> List[str]:
+        """Cluster members hostnames."""
+        if self._peers:
+            return json.loads(self._peers.data[self.app].get("endpoints", "[]"))
+        else:
+            # If the peer relations was not created yet, return only the current member hostname.
+            return [self._endpoint]
+
+    def _update_endpoints(
+        self, endpoint_to_add: str = None, endpoints_to_remove: List[str] = None
```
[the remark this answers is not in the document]
marceloneppel: Great. It's better. Thanks for that. Implemented on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/c7043a0579e3540855a2c33ce5ec297c321b18b2.

src/charm.py
```diff
@@ -68,7 +123,75 @@ def _on_config_changed(self, _):
...
+        if "cluster_initialised" not in self._peers.data[self.app]:
+            return
+
+        try:
+            # Compare set of Patroni cluster members and Juju hosts
+            # to avoid the unnecessary reconfiguration.
+            if self._patroni.cluster_members == self._hosts:
+                return
+
+            logger.info("Reconfiguring cluster")
```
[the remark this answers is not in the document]
marceloneppel: Yes. It makes sense. Implemented on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/c7043a0579e3540855a2c33ce5ec297c321b18b2.

src/charm.py
```diff
@@ -131,14 +294,20 @@ def _patch_pod_labels(self) -> None:
...
             with open("src/resources.yaml") as f:
-                for obj in codecs.load_all_yaml(f, context=self._context):
-                    client.create(obj)
+                for resource in codecs.load_all_yaml(f, context=self._context):
+                    client.create(resource)
+                    logger.info(f"created {str(resource)}")
         except ApiError as e:
-            logger.error("failed to create resources")
-            self.unit.status = BlockedStatus(f"failed to create resources with error {e}")
+            if e.status.code == 409:
```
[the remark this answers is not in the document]
marceloneppel: Moving this change to another PR as it's not related specifically to scale/up and down. Your comment makes total sense. I'm keeping this comment unresolved to add the better explanation in the other PR.

src/charm.py
marceloneppel: Of course. I mixed some try/except blocks here. Thanks for catching this! Fixed on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/eaf5d47b081a0489886e949ee5d4938088cdda6d.

templates/patroni.yml.j2
```diff
@@ -15,13 +15,19 @@ bypass_api_service: true
...
   data_dir: {{ storage_path }}/pgdata
   listen: 0.0.0.0:5432
   pgpass: /tmp/pgpass
+  pg_hba:
+  - host all all 0.0.0.0/0 md5
+  - host replication replication 127.0.0.1/32 md5
+  {%- for endpoint in endpoints %}
+  - host replication replication {{ endpoint }}.{{ namespace }}.svc.cluster.local md5
+  {%- endfor %}
 use_endpoints: true
```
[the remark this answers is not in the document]
marceloneppel: Fixed on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/eaf5d47b081a0489886e949ee5d4938088cdda6d.

src/charm.py
```diff
@@ -68,7 +123,75 @@ def _on_config_changed(self, _):
...
+
+    def add_cluster_member(self, member: str):
+        """Add member to the cluster if all members are already up and running.
+
+        Raises:
+            NotReadyError if either the new member or the current members are not ready.
+        """
+        hostname = self._get_hostname_from_unit(member)
+
+        if not self._patroni.is_all_members_ready():
```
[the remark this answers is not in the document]
marceloneppel: Thanks! Fixed on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/eaf5d47b081a0489886e949ee5d4938088cdda6d.

src/patroni.py
```diff
@@ -61,9 +79,52 @@ def get_postgresql_state(self) -> str:
...
+
+        Returns:
+            True if all members are ready False otherwise. Retries over a period of 10 seconds
+            3 times to allow server time to start up.
+        """
+        # Request info from cluster endpoint
+        # (which returns all members of the cluster and their states).
+        try:
+            for attempt in Retrying(stop=stop_after_delay(10), wait=wait_fixed(3)):
+                with attempt:
```
shayancanonical, line 104: is it better to use tenacity's retry decorator here?

src/charm.py
```diff
@@ -53,22 +63,143 @@ def __init__(self, *args):
...
+        # This also enables the member as part of the cluster.
+        try:
+            self._patch_pod_labels(member)
+        except ApiError as e:
+            logger.error("failed to patch pod")
+            self.unit.status = BlockedStatus(f"failed to patch pod with error {e}")
+            return
+
+    @property
+    def _hosts(self) -> set:
```
shayancanonical, line 183: this property is missing a docstring

shayancanonical (approved):
LGTM, but I have a general comment - the code seems a bit fragmented at times due to the use of helper methods. If at all possible, I would lean towards consolidating all code into one long method over having functionality broken into multiple methods that are harder to trace.

src/patroni.py
MiaAltieri: decorators are the things that go ontop of functions right?

If so, its better to have it nested in the code here since when it is ontop of the function it raises a `RetryError`  when it cannot complete the operation. This function is supposed to return false when the function fails, so this gaurentees it will return false instead of raising a `RetryError`

MiaAltieri (changes requested):

tests/integration/test_charm.py
MiaAltieri: bump

tests/unit/test_charm.py
```diff
@@ -285,7 +298,9 @@ def test_new_password(self):
         self.assertIsNotNone(re.fullmatch("[a-zA-Z0-9\b]{16}$", second_password))
         self.assertNotEqual(second_password, first_password)
 
-    def test_get_postgres_password(self):
+    @patch("charm.PostgresqlOperatorCharm._patch_pod_labels")
+    @patch("charm.PostgresqlOperatorCharm._create_resources")
+    def test_get_postgres_password(self, _, __):
         # Test for a None password.
         self.harness.add_relation(self._peer_relation, self.charm.app.name)
         self.assertIsNone(self.charm._get_postgres_password())
```
MiaAltieri, line 306: Missing tests for these scenarios (i dont believe these are tested in the integration tests which is why I have this comment here, but please correct me if I am wrong on any of these)  : 

[this scenario](https://github.com/canonical/postgresql-k8s-operator/blob/3b9be85fd040d2980e2b571d82039a1636ecb37e/src/charm.py#L79-L82) 
[this](https://github.com/canonical/postgresql-k8s-operator/blob/3b9be85fd040d2980e2b571d82039a1636ecb37e/src/charm.py#L90-L93) 
[this](https://github.com/canonical/postgresql-k8s-operator/blob/3b9be85fd040d2980e2b571d82039a1636ecb37e/src/charm.py#L99-L101)
[this](https://github.com/canonical/postgresql-k8s-operator/blob/3b9be85fd040d2980e2b571d82039a1636ecb37e/src/charm.py#L107-L111)
[this](https://github.com/canonical/postgresql-k8s-operator/blob/3b9be85fd040d2980e2b571d82039a1636ecb37e/src/charm.py#L140-L141)
[this](https://github.com/canonical/postgresql-k8s-operator/blob/3b9be85fd040d2980e2b571d82039a1636ecb37e/src/charm.py#L154-L157)

After looking at the code, there other scenarios that are not tested in the unit tests for patroni.py + charm.py (I was lazy and didn't want to write them out)

But I am happy to write them all out if you'd like, or review your next commit, let me know what you prefer

MiaAltieri (changes requested):

src/patroni.py
marceloneppel: I understood it the way Mia commented. Other approach would be to have two functions. This one with the retry decorator and another one that would do the same that was being done before (checking if a RetryError was thrown and set the return to False, which would be used in the charm code). This approach of only one function seems better to me.

src/charm.py
```diff
@@ -53,22 +63,143 @@ def __init__(self, *args):
...
+        transfer rates in these processes and overload the primary instance.
+        """
+        # Only the leader can reconfigure.
+        if not self.unit.is_leader():
+            return
+
+        # Reconfiguration can be successful only if the cluster is initialised
+        # (i.e. first unit has bootstrap the cluster).
+        if "cluster_initialised" not in self._peers.data[self.app]:
+            return
```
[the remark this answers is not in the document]
marceloneppel: At this point it's not needed because the event will be fired again when the leader sets this flag in the peer relation data bag.

tests/unit/test_charm.py
marceloneppel: Synced offline about implementing them in another PR to not this PR bigger. Ticket DPE-399.

src/charm.py
```diff
@@ -53,22 +63,143 @@ def __init__(self, *args):
...
+
+        # If the leader is the one receiving the event, it adds the new members,
+        # one at a time.
+        if self.unit.is_leader():
+            self._add_members(event)
+
+        # Don't update this member before it's part of the members list.
+        if self._endpoint not in self._endpoints:
+            return
```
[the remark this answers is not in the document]
marceloneppel: It's not needed as the event will be triggered again when the endpoint is added to the endpoints list.

src/charm.py
```diff
@@ -89,56 +235,101 @@ def _on_postgresql_pebble_ready(self, event: WorkloadEvent) -> None:
...
-                # when adding configs to this charm.
-                self._patroni.render_patroni_yml_file()
-                # Restart it and report a new status to Juju.
-                container.restart(self._postgresql_service)
-                logging.info("Restarted postgresql service")
-            # All is well, set an ActiveStatus.
-            self.unit.status = ActiveStatus()
-        else:
+        # Defer the event if pebble is not available yet.
+        if not container.can_connect():
```
paulomach, line 239: Is this possible/needed? I'm assuming that you only get a `pebble_ready` event if the container can connect.

src/charm.py
```diff
@@ -80,6 +211,21 @@ def _on_leader_elected(self, _) -> None:
...
+        # Add this unit to the list of cluster members
+        # (the cluster should start with only this member).
+        if self._endpoint not in self._endpoints:
+            self._add_to_endpoints(self._endpoint)
+
+        # Remove departing units when the leader changes.
+        endpoints_to_remove = self._get_endpoints_to_remove()
+        if endpoints_to_remove:
+            self._remove_from_endpoints(endpoints_to_remove)
+
```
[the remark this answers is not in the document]
marceloneppel: Nice catch. Fixed on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/249f4fda535bf1e54aa824196f0aaefed07a00e3.

src/charm.py
marceloneppel: Thanks. Fixed on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/249f4fda535bf1e54aa824196f0aaefed07a00e3.

src/charm.py
marceloneppel: You're right Paulo. Thanks for catching this. I removed this check on https://github.com/canonical/postgresql-k8s-operator/pull/9/commits/41426336c9ff769597c3990ddcfd931ecb498e48.

src/charm.py
```diff
@@ -53,22 +63,143 @@ def __init__(self, *args):
...
+        if self.unit.is_leader():
+            self._add_members(event)
+
+        # Don't update this member before it's part of the members list.
+        if self._endpoint not in self._endpoints:
+            return
+
+        # Update the list of the cluster members in the replicas to make them know each other.
+        # Update the cluster members in this unit (updating patroni configuration).
+        self._patroni.update_cluster_members()
```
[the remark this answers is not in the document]
marceloneppel: I agree that there'll be some small performance impact of multiple executions when not needed (when for example there is nothing to update in the endpoints, but the relation changed event was triggered by other info that was changed in the relation data bag in the future).

About the other point, related to write Patroni configuration in only one unit, we need to write a configuration file in every unit (it's not a global configuration in the relation or on k8s cluster), so it's needed to have each unit calling this function.

And about the last details, related to having a first relation changed event being executed after a latter one, it seems that the data that is read from the relation comes from the database in the controller and not from a kind of a stored state of the specific event that was deferred as per [this](https://juju.is/docs/sdk/relations#heading--relation-data). But we can confirm that to make sure.

MiaAltieri (approved):
Looks great Marcelo! Thanks for answering all my questions and explaining everything to me! :D

paulomach (approved):
LGTM

src/charm.py
```diff
@@ -188,9 +373,58 @@ def _restart_postgresql_service(self) -> None:
...
+            self._endpoints,
+            self._namespace,
+            self._unit_ip,
+            self._storage_path,
+        )
+
+    @property
+    def _unit_ip(self) -> str:
+        """Current unit ip."""
+        return str(self.model.get_binding(PEER).network.bind_address)
```
jnsgruk, line 389: Is there a reason to be worrying about IPs on Kubernetes when you can use the cluster DNS? I.e. `postgresql-k8s-0.postgresql-k8s-endpoints.model_name.svc.cluster.local`?

marceloneppel: I fixed this now, Jon. I previously had an issue that forced the presence of the IP address only in the Patroni bootstrap process (removing the IP later and using only hostnames). Now this is also fixed (127.0.0.1 is used in the bootstrap process, keeping the change to hostnames after that).

PR #11, opened 2022-06-15
Legacy db and db admin relations

marceloneppel:
# Issue
In short this PR partially address JIRA ticket [DPE-458](https://warthogs.atlassian.net/browse/DPE-458).

In long:
* The new PostgreSQL k8s charm should support legacy relations from the legacy charm.
* There are two relations that should be implemented: db and db-admin.
* These relations should be tested with some charms that already work with the legacy PostgreSQL k8s charm (Mattermost, Discourse and Finos Waltz).
* This is also needed on the new PostgreSQL VM charm.

# Solution
* Implement a library that communicates with PostgreSQL to create users and databases, so we can share this library with the VM charm (and also with the PgBouncer charms, that also have those legacy relations).
* Create some handlers for the legacy relation.

# Context
* In `resources.yaml` some new k8s services are added to map to the primary and the replicas, so even when the primary changes, the application charm can use the same host address to connect to, for example, write some data. It is possible because Patroni updates the pods of the PostgreSQL instances with the correct labels to differentiate primary and replicas.
* In `charm.py` some properties were added to point to the current unit, primary adn replicas endpoints.
* `postgresql.py` is a charm library meant ot be shared between PostgreSQL and PgBouncer charms. It contains some helpers to create users and databases.
* `relations.db.py` has the class that handles both relations, following the same behavior as the legacy charm. It has an additional check for extensions being requested through the relation, which are not allowed. In a later implementation, there should be a mechanism to add extensions through the PostgreSQL charm (using, for example, a config or an action).
* `utils.py` was added to separate code that is used also outside `charm.py`.

# Testing
* New integration test connecting PostgreSQL and Finos Waltz was add to validate the db relation.
* To reduce the size of this PR:
  * Unit tests are part of one of the next three PRs.
  * Testing db relation with Mattermost is part of another of the next three PRs.
  * Testing db-admin relation with Discourse is part of another of the next three PRs.

# Release Notes
* Add legacy db and db-admin relations.


src/charm.py
```diff
@@ -64,6 +71,78 @@ def __init__(self, *args):
         self.framework.observe(self.on.update_status, self._on_update_status)
         self._storage_path = self.meta.storages["pgdata"].location
 
+    def _on_old_db_relation_changed(self, event: RelationChangedEvent) -> None:
```
WRFitch, line 74: It was suggested on another PR that relations could be their own modules sitting in a "relations" subdirectory, in the same way as our other `Relation` objects. I've found this cleans up my charm code, and allows space to independently document the relation. You can also keep test files clean by following the same structure - one file per relation, contained in a "relations" subdir.

src/charm.py
```diff
@@ -64,6 +71,78 @@ def __init__(self, *args):
...
+
+        create_database(connection, database, user)
+
+        logger.error("database created")
+
+        connection.close()
+
+        logger.error("connection closed")
+
+        unit_relation_databag["allowed-subnets"] = ""
```
WRFitch, line 126: Populating the relation databag might get a little messy once it's fully implemented (getting standbys etc) - perhaps this could be extracted to its own method?

src/charm.py
```diff
@@ -64,6 +71,78 @@ def __init__(self, *args):
...
+        connection = connect_to_database(
+            "postgres", "postgres", hostname, self._get_postgres_password()
+        )
+        logger.info(f"Connected to PostgreSQL: {connection}")
+
+        user = f"relation_id_{event.relation.id}_{event.app.name.replace('-', '_')}"
+        password = self._new_password()
+        logger.error(str(event.relation.data))
+        database = event.relation.data[event.app].get("database")
+        if not database:
```
WRFitch, line 102: Would it be worth adding some logs to tell the user about this? If not, could you please add a comment describing it to a reader?

WRFitch (commented):
This is a good start, I'm looking forward to using the postgresql_helpers lib once it's published :)

tests/integration/test_old_relations.py
```diff
@@ -0,0 +1,45 @@
...
+    await asyncio.gather(
+        ops_test.model.deploy(
+            charm, resources=resources, application_name=DATABASE_NAME, trust=True
+        ),
+        ops_test.model.deploy(APPLICATION_NAME, application_name=APPLICATION_NAME),
+    )
+    await ops_test.model.wait_for_idle(apps=[DATABASE_NAME], status="active", timeout=1000)
+
+
+async def test_old_db_relation(ops_test: OpsTest):
```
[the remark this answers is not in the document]
jnsgruk: Yeh, I think `s/old/legacy/g` would be good throughout this PR

lib/charms/postgresql/v0/postgresql_helpers.py
```diff
@@ -0,0 +1,180 @@
...
+
+    Returns:
+         a connection string.
+    """
+    return f"dbname='{database}' user='{user}' host='{host}' password='{password}'"
+
+
+def connect_to_database(
+    database: str, user: str, host: str, password: str
+) -> psycopg2.extensions.connection:
```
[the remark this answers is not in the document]
marceloneppel: Good suggestion! Added it.

src/relations/db.py
```diff
@@ -0,0 +1,258 @@
...
+            "cluster_initialised" not in self.charm._peers.data[self.charm.app]
+            or not self.charm._patroni.member_started
+        ):
+            event.defer()
+            return
+
+        if not self.charm.unit.is_leader():
+            return
+
+        logger.warning(f"DEPRECATION WARNING - `{event.relation.name}` is a legacy interface")
```
WRFitch, line 84: I find `{self.relation_name}` a little easier to parse, but it's a matter of personal preference.

src/relations/db.py
```diff
@@ -0,0 +1,258 @@
...
+
+        unit_relation_databag = event.relation.data[self.charm.unit]
+        application_relation_databag = event.relation.data[self.charm.app]
+
+        # Do not allow apps requesting extensions to be installed.
+        if "extensions" in unit_relation_databag or "extensions" in application_relation_databag:
```
WRFitch, line 90: I found some code to install extensions in the old pgbouncer charm. Do you want me to reject extensions in a similar way in the new charm, or pass them on into the postgres relation?

src/relations/db.py
```diff
@@ -0,0 +1,258 @@
...
+                    dbname=database,
+                    port=5432,
+                    user=user,
+                    password=password,
+                    fallback_application_name=event.app.name,
+                )
+            )
+
+            # Set the data in both application and unit data bag.
+            # It 's needed to run this logic on every relation changed event
```
WRFitch, line 139:
```suggestion
            # It's needed to run this logic on every relation changed event
```

src/relations/db.py
```diff
@@ -0,0 +1,258 @@
...
+                databag["allowed-units"] = self._get_allowed_units(event.relation)
+                databag["host"] = self.charm.endpoint
+                databag["master"] = primary
+                databag["port"] = "5432"
+                databag["standbys"] = standbys
+                databag["state"] = self._get_state()
+                databag["version"] = self.charm.postgresql.get_postgresql_version()
+                databag["user"] = user
+                databag["password"] = password
+                databag["database"] = database
```
WRFitch, line 156: Shayan mentioned a slightly cleaner way of doing this in the pgbouncer charm:
```python
databag = {
    "allowed-subnets": self._get_allowed_subnets(event.relation),
    "allowed-units": self._get_allowed_units(event.relation),
    # ....
}

databag.update(updates)
```

tests/integration/helpers.py
```diff
@@ -1,11 +1,80 @@
...
+import psycopg2
 import requests
+import yaml
 from pytest_operator.plugin import OpsTest
 
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+DATABASE_APP_NAME = METADATA["name"]
+
+
+async def check_database_users_existence(
```
WRFitch, line 17: These should be super useful for pgbouncer integration tests

WRFitch (approved):
Good stuff Marcelo, I got distracted halfway through reviewing because I was implementing your postgresql_helpers lib and it's already super useful.

src/relations/db.py
marceloneppel: Yes. For now we should also reject extensions on PgBouncer in a similar way.

src/relations/db.py
marceloneppel: Nice suggestion. I implemented it on https://github.com/canonical/postgresql-k8s-operator/pull/11/commits/33c493834552bc49d9c9d3b06a58474c87e61c7e.

src/relations/db.py
marceloneppel: Good catch. I forgot about this attribute. I changed it on https://github.com/canonical/postgresql-k8s-operator/pull/11/commits/33c493834552bc49d9c9d3b06a58474c87e61c7e.

paulomach (approved):
LGTM

.github/workflows/ci.yaml
```diff
@@ -65,6 +65,8 @@ jobs:
         uses: charmed-kubernetes/actions-operator@main
         with:
           provider: microk8s
+          # This is needed until https://bugs.launchpad.net/juju/+bug/1977582 is fixed.
+          bootstrap-options: "--agent-version 2.9.29"
```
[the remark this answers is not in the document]
marceloneppel: Yes, it is required for that. There is also another issue related to scale down that needs to be fixed on Juju to implement a better scale down logic on the relation departed event.

PR #12, opened 2022-06-23
Implement new relation

marceloneppel:
# Issue
In short this PR partially address JIRA ticket [DPE-463](https://warthogs.atlassian.net/browse/DPE-463).

In long:
* The new PostgreSQL k8s charm should support a new relation following the pattern from [this relation interface](https://github.com/canonical/charm-relation-interfaces/tree/main/database).

# Solution
* Implement a new relation using the [data platform provides library](https://github.com/canonical/data-platform-libs/blob/main/lib/charms/data_platform_libs/v0/database_provides.py).

# Context
* Files copied from [Data Platform Libs](https://github.com/canonical/data-platform-libs) repository that don't need to be reviewed:
  * `lib/charms/data_platform_libs/v0/database_provides.py`
  * `tests/integration/new_relations/application-charm/*`
* In `lib/charms/postgresql_k8s/v0/postgresql.py`, an extra flag was added to enable the `extra user roles` field to be used during the database user creation.
* In `src/relations/postgresql_client.py` the relation handlers are implemented. There is a `on database requested` handler instead of the often used `on relation changed` handler. It is using the Data Platform Libs provides library. There is also a `on relation broken` handler that removes the previously created database user when the relation is gone.
* In `src/charm.py`, the Pebble layer `PATRONI_SCOPE` environment variable was changed to have a prefix and use the app name instead of the namespace because, the way it was before, multiple clusters deployed in the same model would face conflicts in the leader election and all of the units would become replicas (this was figured out when deploying multiple clusters in the integration tests of this PR). Also a prefix (patroni) was added to avoid conflict with the services that are automatically created by charmed operators (which otherwise would cause the same issue).
* The cluster name was added as a label selector to `src/resources.yaml` so multiple PostgreSQL clusters can be deployed in the same model without facing conflicts in the service names.

# Testing
* Unit tests for the postgresql client relation file will be added in the next PR to reduce the size of this PR.
* Existing testes were fixed to check for the correct cluster name.
* Integration tests for the new relation (`tests/integration/new_relations/test_new_relations.py`) were copied from [data platform libs integration tests](https://github.com/canonical/data-platform-libs/blob/main/tests/integration/test_charm.py) with some minor changes:
  * In this repository the PostgreSQL version number is retrieved instead of the complete version string.
  * One additional check was made by connecting to a read replica and trying to alter some data.

# Release Notes
* Add new  PostgreSQL relation.


tests/integration/new_relations/helpers.py
```diff
@@ -42,8 +42,34 @@ async def build_connection_string(
...
+                "service",
+                "-n",
+                ops_test.model_name,
+                host.split(".")[0],
+                "-o",
+                "json",
+                "|",
+                "jq",
+                "-r",
+                ".spec.clusterIP",
```
jnsgruk, line 63: Consider using lightkube here to fetch the service so that you don't accidentally tie our tests to microk8s?

marceloneppel: Thanks Jon. Good suggestion. I will update it.

marceloneppel: Implemented on https://github.com/canonical/postgresql-k8s-operator/pull/12/commits/70d1347458d5e7e2792f0c4447d331ff6bdb2047.

src/relations/postgresql_client.py
```diff
@@ -0,0 +1,135 @@
...
+            self.charm.postgresql.create_user(user, password, extra_user_roles=extra_user_roles)
+            self.charm.postgresql.create_database(database, user)
+
+            # Share the credentials with the application.
+            self.database_provides.set_credentials(event.relation.id, user, password)
+
+            # Set the read/write endpoint.
+            self.database_provides.set_endpoints(
+                event.relation.id,
+                f"{self.charm.primary_endpoint}:5432",
```
MiaAltieri, line 91: nit: make 5432 a constant

src/relations/postgresql_client.py
```diff
@@ -0,0 +1,135 @@
...
+            )
+
+            # Set the read-only endpoint only if there are replicas.
+            if len(self.charm._peers.units) > 0:
```
MiaAltieri, line 95: Question: If a replica gets added later on will the `DatabaseRequestedEvent ` get triggered and will this be called?

src/relations/postgresql_client.py
```diff
@@ -0,0 +1,135 @@
+# Copyright 2022 Canonical Ltd.
```
MiaAltieri, line 1: I think this file should be changed to `postgresql_provider.py` since this is provider code. (Also mongodb follows this schema and it would be nice if they followed the same pattern.)

src/relations/postgresql_client.py
marceloneppel: Good point Mia. It was not being handled. I fixed that adding calls to update it on peer relation changed and departed events, and also two new integration tests.

Implemented on https://github.com/canonical/postgresql-k8s-operator/pull/12/commits/79fc87d828b1d6bb996c2055060ed37b2afc55a1.

marceloneppel: The method from the above commit was improved in https://github.com/canonical/postgresql-k8s-operator/pull/12/commits/188021226e6bec2ef52fb042881ce83ac1ad982e.

WRFitch (approved):
LGTM!

MiaAltieri (approved):
LATM (looks awesome to me)

PR #13, opened 2022-06-30
Remove k8s resources on application removal

marceloneppel:
# Issue
* Both the PostgreSQL k8s charm and the workload it runs (Patroni) create some k8s resources that are not removed when the application is removed from a model.
* This causes some issues to a new deployment of the charm in the same model later (Patroni can't elect a leader for the cluster).

# Solution
* During the charm stop event (application removal), remove the resources created by the PostgreSQL k8s charm and also the resources that were created by Patroni.

# Context
* In `charm.py`, the on stop event removes the k8s resources created in on install and also by Patroni.
* In the on stop hook the leader unit was not checked because the leadership changes during the removal and sometimes if it is checked before we can do the resources removal we end up removing nothing.
* In `resources.yaml` two unneeded resources were removed.
* Some helpers were created for the integration tests.
* Integration tests were added for checking that the right resources are created and then later removed.
* Some details in the tests will be improved in the future, specially the charm build process (we should a fixture, and also update the other tests that are building the charm).

# Testing
* New integration tests were implemented to test that the correct resources were created when the application is deployed and also to test that all of them are removed when the application is removed from the model.
* There is also a new test that checks that it works after being removed and deployed again in the same model.

# Release Notes
* Remove k8s resources created by the PostgreSQL k8s charm and Patroni when the application is removed.


WRFitch (approved):

shayancanonical (approved):
lgtm!

Issue #3, opened 2022-01-24
Fix automatic failover handling


marceloneppel:
Currently, the automatic failover of Patroni should be handled with a restart of the service being called in the `update status hook` until we have some sort of signal forwarding mechanism in pebble, because when an automatic failover is triggered (because the leader Postgres instance got some problem, for example) Patroni waits indefinitely for the Postgres process to stop in the replica instances, but currently we have a Postgres zombie process with is not aknowledged by Patroni because of a lack of signal forwarding.

This signal forwarding mechanism is being discussed in https://github.com/canonical/pebble/issues/6.

marceloneppel:
The same issue (zombie process) affects the manual failover and switchover actions.

For all this cases, we are executing health checks through the Patroni REST API and restarting the Postgres+Patroni service if it keeps stuck for some time.

PR #14, opened 2022-07-01
Remove Patroni stop/restart workaround

marceloneppel:
# Issue
In short this PR address https://github.com/canonical/postgresql-k8s-operator/issues/3.

In long:
* After https://github.com/canonical/pebble/issues/6 was resolved, we needed to remove a previous workaround that was implemented in this charm to handle the situation when a failover happens and makes Patroni restart the PostgreSQL process inside a unit. This was not working before the mentioned issue was solved (the process would become stuck forever, preventing the operation to finish).

# Solution
* Remove all the code related to the workaround that was created in the past in this charm.

# Context
* This PR is focused in removing code that is not needed anymore.
* Only the comment in `tests/integration/test_charm.py` was removed. The change from `update status config` to `fast forward` should be done in the future, to avoid a larger diff in this PR.

# Testing
* It was tested manually.
* Also, the existing integrations tests should continue to work with this change (that is the way we confirm that the bug is fixed for the combination of Patroni + Pebble: the scale down works correctly, without stuck Patroni/PostgreSQL processes in the units preventing a failover when the leader is removed or have some issue).

# Release Notes
* Remove workaround for https://github.com/canonical/pebble/issues/6.


shayancanonical (approved):
:scissors:

paulomach (approved):
this kind of changes are so satisfying

PR #17, opened 2022-07-06
Add Mattermost test for the db legacy relation

marceloneppel:
# Issue
In short this PR partially address JIRA ticket [DPE-458](https://warthogs.atlassian.net/browse/DPE-458).

In long:
* The legacy db relation in the PostgreSQL charm should be tested with Mattermost. 

# Solution
* Add a test similar to the one that was created using Finos Waltz on https://github.com/canonical/postgresql-k8s-operator/pull/11.

# Context
* In `charm.py` a temporary implementation of TLS was added. It is needed because Mattermost needs TLS enabled when connecting to PostgreSQL. The idea at this time to make the relation work is to import a generated key and a certificate using Juju resources.
* In `metadata.yaml` two new resources were added for the temporary TLS implementation.
* In `tests/gen-test-certs.sh` a script was created to generate the TLS resources for the integration test.
* In `tox.ini` the above script is called before the integrations tests.
* In `tests/integration/test_db_with_tls.py` the test using Mattermost was implemented.

# Testing
* New integration test connecting PostgreSQL and Mattermost was added to validate the db relation.

# Release Notes
* Add legacy db relation test using Mattermost.


metadata.yaml
```diff
@@ -20,6 +20,12 @@ resources:
     type: oci-image
     description: OCI image for PostgreSQL (postgresql-patroni)
     upstream-source: postgresql-patroni
+  cert-file:
+    type: file
+    filename: server.crt
+  key-file:
+    type: file
+    filename: server.key
```
jnsgruk, line 28: @delgod @marceloneppel - I think we're gonna avoid having these resources in favour of using the `tls-certificates-operator` and a relation?

marceloneppel: Yeah Jon. I am not exactly sure about the right implementation, but this one is just temporary to ensure the legacy relation with Mattermost works. There is a deprecation warning being logged when these resources are attached to the charm.

jnsgruk: In that case, I'd prefer to land this test. Adding functionality to the charm that we know is shortly going to be "unsupported" feels like a flawed idea to me. I'd prefer that we build the charm from [this OR](https://github.com/canonical/tls-certificates-operator/pull/1) manually in the test and implement the tls-certificates relation in

lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -0,0 +1,170 @@
...
+                provided when the object for this class was created).
+
+        Returns:
+             psycopg2 connection object.
+        """
+        connection = psycopg2.connect(
+            f"dbname='{database if database else self.database}' user='{self.user}' host='{self.host}' password='{self.password}' connect_timeout=1"
+        )
+        connection.autocommit = True
+        return connection
```
marcoppenheimer, line 85: I'm a PG noob, so it's probably fine. 

Do you need to `close()` or `stop()` `cursor` or `connection` instances?

src/relations/db.py
```diff
@@ -0,0 +1,259 @@
...
+            self.charm.unit.status = BlockedStatus("extensions requested through relation")
+            return
+
+        # Sometimes a relation changed event is triggered,
+        # and it doesn't have a database name in it.
+        database = event.relation.data[event.app].get("database")
+        if not database:
+            logger.warning("No database name provided")
+            event.defer()
+            return
```
marcoppenheimer, line 104: In what cases would a `defer()` resolve a missing `database` key from the relation data?
Do you expect it to happen eventually, set from the provider lib?

src/relations/db.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def _get_allowed_subnets(self, relation: Relation) -> str:
+        """Build the list of allowed subnets as in the legacy charm."""
+
+        def _csplit(s) -> Iterable[str]:
+            if s:
+                for b in s.split(","):
+                    b = b.strip()
+                    if b:
+                        yield b
```
marcoppenheimer, line 229: As this is called (I think...) only once, could this be part of the logic in `_get_allowed_subnets`?

tests/unit/test_charm.py
```diff
@@ -300,3 +290,59 @@ def test_get_postgres_password(self, _, __):
...
+            self.assertEqual(_update_config.call_count, 2)
+
+        # Test with both resources.
+        with patch(
+            "charm.PostgresqlOperatorCharm._tls_files",
+            PropertyMock(return_value=[Path("/tmp/server.crt"), Path("/tmp/server.key")]),
+        ):
+            self.charm._store_tls_files()
+            self.assertEqual(_push.call_count, 2)
+            self.assertEqual(_update_config.call_count, 3)
```
marcoppenheimer, line 331: More of a choice than a comment. 

What do you think about splitting these tests out into 3 separate tests? Same with some of the other units. 
It does add boiler plate, but it might be easier to specify which exact behavior fails during a test failure.
It might also mean we don't have to keep an internal track of `_update_config.call_count` and update that when we modify this test. 

Though it's a nitpick, definitely not important.

marcoppenheimer (approved):
Overall I think it looks good! Minor nits.

lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def delete_user(self, user: str) -> None:
+        """Deletes a database user.
+
+        Args:
+            user: user to be deleted.
+        """
+        # List all databases.
+        try:
+            with self._connect_to_database() as connection, connection.cursor() as cursor:
```
shayancanonical, line 137: is there a reason why there are 3 calls to `self._connect_to_database` in this method? is to break up the method logic into 3 transactions?

src/relations/db.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def _get_state(self) -> str:
+        """Gets the given state for this unit.
+
+        Returns:
+            The state of this unit. Can be 'standalone', 'master', or 'standby'.
+        """
+        if len(self.charm._peers.units) == 0:
+            return "standalone"
+        if self.charm._patroni.get_primary(unit_name_pattern=True) == self.charm.unit.name:
```
shayancanonical, line 256: minor nit: this could be an `elif`

tests/gen-test-certs.sh
```diff
@@ -0,0 +1,14 @@
...
+chmod og-rwx tests/tls/root.key
+openssl x509 -req -in tests/tls/root.csr -text -days 3650 \
+  -extfile /etc/ssl/openssl.cnf -extensions v3_ca \
+  -signkey tests/tls/root.key -out tests/tls/root.crt
+openssl req -new -nodes -text -out tests/tls/server.csr \
+  -keyout tests/tls/server.key -subj "/CN=dbhost.example.com"
+chmod og-rwx tests/tls/server.key
+openssl x509 -req -in tests/tls/server.csr -text -days 365 \
+  -CA tests/tls/root.crt -CAkey tests/tls/root.key -CAcreateserial \
+  -out tests/tls/server.crt
```
shayancanonical, line 14: minor nit: missing new line

shayancanonical (approved):
LGTM, just some minor nits

PR #22, opened 2022-07-28
Rename relation interface names

marceloneppel:
# Issue
Relation interface name must not include hyphens.

# Solution
Change `-` to `_` in the relation interface names.

# Context
* Fixed `postgresql-replicas` relation, which has the incorrect name and also the interface name in the wrong format.
* Also removed the duplicate CI run on `ci.yaml` when the code is merged to main. This was done because there is already a workflow running through `release.yaml`.
* Also fixed the juju python library in `tox.ini`. The dependecy was pointing to git instead of the published version, which caused some issues in the integration tests because of some breaking changes (one of them, related to https://github.com/juju/python-libjuju/commit/d42f7e6d6095c0ff7ed807764779b395092e2f7e, was fixed now on `tests/integration/test_charm.py` to avoid errors again when a new version of the library is published on PyPi).

# Release Notes
* Rename relation interface names.
* Remove duplicate CI run when changed are merged to main.
* Change juju dependency in tox.ini to the one that is published.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,8 +1,5 @@
 name: Tests
 on:
-  push:
-    branches:
-      - main
   pull_request:
 
 jobs:
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -22,12 +22,12 @@ resources:
     upstream-source: dataplatformoci/postgres-patroni
 
 peers:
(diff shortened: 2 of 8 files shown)


MiaAltieri (approved):
LGTM!! :D

WRFitch (approved):

PR #23, opened 2022-08-03
Update Charmhub action

marceloneppel:
Update Charmhub action to 2.0.0-rc to avoid creating revisions named `revundefined`.

* Changed flake version because the new version is not compatible with the plugin that is used in the CI.


MiaAltieri (approved):
stellar!

WRFitch (approved):

PR #25, opened 2022-08-16
Change charm database user

marceloneppel:
# Issue
* Database charms need a database user with a name that is easy to identify. This user shouldnt be used by charms related to the database charm.

# Solution
* Change the default `postgres` superuser to a user called `operator`.

# Context
* Changed the `postgres` superuser to `operator`. Patroni uses that for some operations too.
* Patroni uses another user for replication, so no issues because of changing the superuser. 

# Testing
* Unit and integration tests were updated.
* Manually tested the charm to check that the right user is created inside PostgreSQL.

# Release Notes
* Change charm user to a user called `operator`.

### DIFF
--- a/actions.yaml
+++ b/actions.yaml
@@ -3,5 +3,6 @@
 
 get-primary:
   description: Get the unit with is the primary/leader in the replication.
-get-postgres-password:
-  description: Get the initial postgres user password for the database.
+get-operator-password:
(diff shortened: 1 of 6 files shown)


MiaAltieri (approved):
😎

zmraul (approved):

PR #27, opened 2022-08-23
Rework secrets

marceloneppel:
# Issue
* Secrets like passwords and TLS certificates should be moved to Juju secrets in the future.
* Currently, passwords are being set and retrieved from app relation data in multiple places, what can lead to a big refactor later (even more when we add TLS certificates to the charm).

# Solution
* Move secret implementation logic to specific functions that can be easily refactored in the future to use Juju secrets instead of app relation data.

# Context
* Unit peer relation property was also added for future use cases.
* Some constants were created to avoid the same strings in multiple places.

# Testing
* Additional unit tests were added.
* Tested using integration tests (to ensure the new implementation works the same way).

# Release Notes
* Add specific functions to set and get secrets.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -30,7 +30,7 @@ from ops.pebble import Layer
 from requests import ConnectionError
 from tenacity import RetryError
 
-from constants import PEER, USER
+from constants import PEER, REPLICATION_PASSWORD_KEY, USER, USER_PASSWORD_KEY
 from patroni import NotReadyError, Patroni
 from relations.db import DbProvides
 from relations.postgresql_provider import PostgreSQLProvider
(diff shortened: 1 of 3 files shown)


paulomach (approved):
no comments, pristine

src/charm.py
```diff
@@ -69,6 +69,48 @@ def __init__(self, *args):
         self.legacy_db_relation = DbProvides(self, admin=False)
         self.legacy_db_admin_relation = DbProvides(self, admin=True)
 
+    @property
+    def app_peer_data(self) -> Dict:
```
WRFitch, line 73: You might want to consider a separate peers.py class in the relations directory in future, since it's technically a relation.

WRFitch (approved):

src/charm.py
marceloneppel: Yeah, good idea. In this specific case, it should be changed to use Juju secrets in the future, so I think it can be kept here.

PR #29, opened 2022-08-26
PostgreSQL TLS implementation

marceloneppel:


templates/patroni.yml.j2
```diff
@@ -23,6 +23,12 @@ postgresql:
   custom_conf: {{ storage_path }}/postgresql-k8s-operator.conf
   data_dir: {{ storage_path }}/pgdata
   listen: 0.0.0.0:5432
+  {%- if enable_tls %}
+  parameters:
+    ssl: on
+    ssl_cert_file: {{ storage_path }}/external-ca.crt
+    ssl_key_file: {{ storage_path }}/external-cert.pem
+  {%- endif %}
```
delgod, line 31: I think we should also set `ssl_ca_file` here, which should contain CA + chain in one file.

PR #35, opened 2022-09-06
TLS relations files

marceloneppel:
# Issue
* PostgreSQL charm needs to have a way to enable TLS in the connections between instances and from the client applications to the PostgreSQL server.

# Solution
* This PR starts to solve that issue by adding some files needed to interact with https://github.com/canonical/tls-certificates-operator (to later, in the [next PR](https://github.com/canonical/postgresql-k8s-operator/pull/36), request TLS certificates and load them).

# Context
* The larger diff presented by the number of changed lines doesn't reflect what needs to be reviewed (which is a small part of the changes of this PR).
* Files that were `IMPORTED` and can be `IGNORED`:
  * lib/charms/rolling_ops/v0/rollingops.py
  * lib/charms/tls_certificates_interface/v1/tls_certificates.py
* You can `FOCUS` your review on:
  * `lib/charms/postgresql_k8s/v0/postgresql_tls.py`
* charmcraft.yaml and requirements.txt only contains dependencies needed for lib/charms/tls_certificates_interface/v1/tls_certificates.py.
* metadata.yaml contains two new relations that are needed (one for triggering rolling restarts after TLS is enabled/disabled and the other to request/renew TLS certificates).

# Testing
* Unit tests for `lib/charms/postgresql_k8s/v0/postgresql_tls.py` were added in the [next PR](https://github.com/canonical/postgresql-k8s-operator/pull/36) to avoid a larger diff here due to changes that are needed on `_get_secret` and `_set_secret` methods.
* Integration tests were added in the [third PR](https://github.com/canonical/postgresql-k8s-operator/pull/37) of this sequence.

# Release Notes
* Add TLS relations files (to handle TLS certificates request and rolling restarts).

PR #36, opened 2022-09-06
TLS implementation

marceloneppel:
# Issue
* PostgreSQL charm needs to have a way to enable TLS in the connections between instances and from the client applications to the PostgreSQL server.

# Solution
* This PR continues to solve that issue by implementing the interaction with https://github.com/canonical/tls-certificates-operator using lib/charms/postgresql_k8s/v0/postgresql_tls.py and uploading certificates to the PostgreSQL container to enable TLS. It doesn't take care of enabling TLS on Patroni API, only on PostgreSQL connections.

# Context
* This PR should be merged only after https://github.com/canonical/postgresql-k8s-operator/pull/35.
* Files that you can skip during review:
  * tests/unit/test_charm.py (because it changed mostly because of a refactor on method names)
* You can `FOCUS` your review on:
  * `src/charm.py` (ignoring the changes to `get_secret` and `set_secret` methods, which was only a refactor needed to use them outside the class, like in lib/charms/postgresql_k8s/v0/postgresql_tls.py; FOCUS on the other methods, that push TLS files to the container and later enable/disable TLS, also performing a rolling restart, which is needed for the other instances to start to use TLS in the replication connection)
  * `templates/patroni.yml.j2` (changes to enable/disable SSL in the client and other instances connections; this is something specific to PostgreSQL/Patroni)

# Testing
* Unit tests were updated on `tests/unit/test_charm.py` and `tests/unit/test_patroni.py`.
* Unit tests were added for from the [previous PR](https://github.com/canonical/postgresql-k8s-operator/pull/35) through `tests/unit/test_postgresql_tls.py`.
* Integration tests were added in the [next PR](https://github.com/canonical/postgresql-k8s-operator/pull/37).

# Release Notes
* Add TLS implementation.


src/charm.py
```diff
@@ -649,6 +676,70 @@ def _peers(self) -> Relation:
...
+        except RetryError as e:
+            logger.error("failed to restart PostgreSQL")
+            self.unit.status = BlockedStatus(f"failed to restart PostgreSQL with error {e}")
+
+    def update_config(self) -> None:
+        """Updates Patroni config file based on the existence of the TLS files."""
+        enable_tls = False
+        key, ca, cert = self.tls.get_tls_files()
+        if None not in [key, ca, cert]:
+            enable_tls = True
```
marcoppenheimer, line 728: nit, but I wonder if it's more explicit having a:

```python
enable_tls = all(self.tls.get_tls_files())
```

tests/unit/test_postgresql_tls.py
```diff
@@ -0,0 +1,200 @@
...
+        self.assertIsNotNone(self.charm.get_secret(SCOPE, "csr"))
+        _request_certificate_creation.assert_called_once()
+
+    def test_parse_tls_file(self):
+        # Test with a plain text key.
+        key = self.get_content_from_file(filename="tests/unit/key.pem")
+        parsed_key = self.charm.tls._parse_tls_file(key)
+        self.assertEqual(parsed_key, key.encode("utf-8"))
+
+        # Test with a base4 encoded key.
```
WRFitch, line 120:
```suggestion
        # Test with a base64 encoded key.
```

src/charm.py
```diff
@@ -649,6 +676,70 @@ def _peers(self) -> Relation:
...
+
+        # Update and reload configuration based on TLS files availability.
+        self._patroni.render_patroni_yml_file(enable_tls=enable_tls)
+        if not self._patroni.member_started:
+            return
+
+        restart_postgresql = enable_tls != self.postgresql.is_tls_enabled()
```
marcoppenheimer, line 735: nit, I think it would be more obviously parsable to have boring `if`'s here.

```python
if self.postgresql.is_tls_enabled() != enable_tls:
	self.on[self.restart_manager.name].acquire_lock.emit()
```

WRFitch (approved):

src/charm.py
marceloneppel: Good catch! I will change it here. Thanks Marc!

src/charm.py
marceloneppel: In this specific part of the code, the check must be made before `self._patroni.reload_patroni_configuration()` is called (because it enables TLS, which is a step needed before restarting PostgreSQL).

`self._patroni.reload_patroni_configuration()` is called before the if because it is also needed when other configuration change.

Is it ok to keep the check this way due to that needs?

WRFitch (approved):

src/charm.py
marceloneppel: Updated on https://github.com/canonical/postgresql-k8s-operator/pull/36/commits/9e4d48110290a0002771b41ced72413d2d227dd1.

PR #38, opened 2022-09-06
TLS on Patroni REST API

marceloneppel:
# Issue
* Both Patroni REST API (that is used by this charm for some operations, and also by Patroni itself) and the patronictl command line tool (tool used by Patroni for some operations; it uses Patroni REST API under the hood) need to have the communication to and from them being encrypted.

# Solution
* Enable TLS on both Patroni REST API and patronictl command line tool when TLS is enabled on PostgreSQL (what happens after a relations is established with the TLS Certificates Operator).

# Context
* When TLS is enabled, the non encrypted endpoint is not available anymore; the opposite happens when TLS is disabled: the encrypted endpoint is not available anymore. So, with the current implementation, even withouth an integration test, the correct implementation of TLS on the API is validated (because the API is used in several events of the charm and any wrong call would make the charm not deploy correctly).
* Summary of the main changes (you can `FOCUS` your review on them):
  * `src/patroni.py`: Patroni URL is now built based on TLS enabled/disabled. Also, every call uses the CA bundle file to verify the certificate from the server.
  * `templates/patroni.yml.j2`: TLS configuration (enabled when TLS is also enabled on PostgreSQL).
  * `tests/integration/helpers.py` and `tests/integration/test_tls.py`: new checks to ensure the API is accepting TLS connections.

# Testing
* The integration test for TLS was updated to check that Patroni REST API has TLS enabled.

# Release Notes
* Add TLS encryption to Patroni REST API and patronictl command line tool.


lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -178,19 +186,25 @@ def get_postgresql_version(self) -> str:
...
+            with self._connect_to_database(
+                connect_to_current_host=check_current_host
+            ) as connection, connection.cursor() as cursor:
                 cursor.execute("SHOW ssl;")
                 return "on" in cursor.fetchone()[0]
-        except psycopg2.Error as e:
-            logger.error(f"Failed to get check whether TLS is enabled: {e}")
-        return False
+        except psycopg2.Error:
+            # Connection errors happen when PostgreSQL has not started yet.
```
marcoppenheimer, line 206: Nit, maybe debug log `e.output`

lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -178,19 +186,25 @@ def get_postgresql_version(self) -> str:
             logger.error(f"Failed to get PostgreSQL version: {e}")
             raise PostgreSQLGetPostgreSQLVersionError()
 
-    def is_tls_enabled(self) -> bool:
+    def is_tls_enabled(self, check_current_host: bool = False) -> bool:
```
marcoppenheimer, line 189: Maybe it's not appropriate, so feel free to ignore.

Is it valuable to, instead of taking `check_current_host`, instead to take a host address?
That way you could arbitrarily check address for TLS, or pass `self.charm.unit.private_address` or whatever. Not sure if that's useful/appropriate for this use-case though.

src/patroni.py
```diff
@@ -48,6 +51,14 @@ def __init__(
         self._planned_units = planned_units
         self._superuser_password = superuser_password
         self._replication_password = replication_password
+        self._tls_enabled = tls_enabled
+        # Variable mapping to requests library verify parameter.
+        self._verify = f"{self._storage_path}/{TLS_CA_FILE}" if tls_enabled else True
```
marcoppenheimer, line 56: If I understand correctly, `requests.get(verify=True)` using the local public root certificate. Where is that in this case?

marcoppenheimer: Shouldn't it be `verify=False if not tls_enabled`?

marcoppenheimer (dismissed):

lib/charms/postgresql_k8s/v0/postgresql.py
marceloneppel: Fixed on https://github.com/canonical/postgresql-k8s-operator/pull/38/commits/9ff5c810f464af161bc0345696f48226ee381ce5.

src/patroni.py
marceloneppel: I added a better comment on https://github.com/canonical/postgresql-k8s-operator/pull/38/commits/9ff5c810f464af161bc0345696f48226ee381ce5. The `verify` flag is set to the CA bundle file when TLS is enabled to correctly verify the API server certificates. When TLS is disabled, the charm set the True value (which Is the [default](https://github.com/psf/requests/blob/2a6f290bc09324406708a4d404a88a45d848ddf9/requests/api.py#L39) from the library), which seems to not validate anything when TLS is disabled (and we are requesting something from an http url).

lib/charms/postgresql_k8s/v0/postgresql.py
marceloneppel: In this case, I don't see the need for checking in that way. It seems to be good, if we find a use case that needs that. Probably in the future this method may be removed to a better check inside the charm code.

marcoppenheimer (approved):

WRFitch (approved):

PR #45, opened 2022-09-22
Add initial backup and rewind settings

marceloneppel:
# Issue
* Jira issue: [DPE-700](https://warthogs.atlassian.net/browse/DPE-700)
* PostgreSQL needs some initial settings for backups and rewind operations (synchronisation of the data between a primary and the replicas after a replica is promoted to primary).
* We also need to ensure that a rewind operations (the ones using the [pg_rewind](https://www.postgresql.org/docs/current/app-pgrewind.html) tool) are done using connections encrypted by TLS.

# Solution
* Add the following settings for backup operations:

  * achieve_mode=on

  * archive_command=/bin/true

  * wal_level = logical

* Add the following settings for rewind operations:

  * remove_data_directory_on_rewind_failure = true

  * remove_data_directory_on_diverged_timelines = true

Add a user for pg_rewind and test that it uses TLS when connecting to other PostgreSQL instance.

# Context
* pg_rewind had a new user created only for it, and that user was added in the list of users that can have the password rotated.

* You can `FOCUS` more on the following files when reviewing the code:

  * `templates/patroni.yml.j2`: contains the new backup and rewind settings.

  * `tests/integration/helpers.py`: container new helper methods used to ensure that pg_rewind is using TLS in its connections.

  * `tests/integration/test_charm.py`: the integration test was updated to ensure that the new settings are the current settings in the database.

  * `tests/integration/test_tls.py`: adds an additional check that ensures pg_rewind is using TLS in its connections.

* Other files are related to the new pg_rewind user.

# Testing
* Existing unit tests on `tests/unit/test_patroni.py` were updated to match the new settings.
* The extra backup and rewind settings are now added to the integration test that checks the current instance configuration.
* An additional check was added to ensure that `pg_rewind` is using TLS.

# Release Notes
* Add initial settings for backup and rewind operations.
* Add user for pg_rewind.
* Add test to ensure pg_rewind is using TLS.

### DIFF
--- a/templates/patroni.yml.j2
+++ b/templates/patroni.yml.j2
@@ -2,6 +2,9 @@ bootstrap:
   dcs:
     postgresql:
       use_pg_rewind: true
+      parameters:
+        max_wal_senders: 3
+        wal_level: logical
   initdb:
   - auth-host: md5
   - auth-local: trust
--- a/templates/patroni.yml.j2
+++ b/templates/patroni.yml.j2
@@ -58,8 +58,8 @@ postgresql:
     replication:
       password: {{ replication_password }}
     rewind:
-      username: {{ backup_user }}
-      password: {{ backup_password }}
+      username: {{ rewind_user }}
+      password: {{ rewind_password }}
     superuser:
       password: {{ superuser_password }}
 use_endpoints: true
--- a/templates/patroni.yml.j2
+++ b/templates/patroni.yml.j2
@@ -7,8 +7,6 @@ bootstrap:
       parameters:
         archive_command: /bin/true
         archive_mode: on
-        max_replication_slots: {{ max_wal_senders }}
-        max_wal_senders: {{ max_wal_senders }}
         wal_level: logical
   initdb:
   - auth-host: md5
--- a/templates/patroni.yml.j2
+++ b/templates/patroni.yml.j2
@@ -2,8 +2,12 @@ bootstrap:
   dcs:
     postgresql:
       use_pg_rewind: true
+      remove_data_directory_on_rewind_failure: true
+      remove_data_directory_on_diverged_timelines: true
       parameters:
-        max_wal_senders: 3
+        archive_command: /bin/true
+        archive_mode: on
+        max_wal_senders: {{ max_wal_senders }}
         wal_level: logical
   initdb:
   - auth-host: md5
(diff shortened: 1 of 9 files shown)


templates/patroni.yml.j2
```diff
@@ -2,6 +2,12 @@ bootstrap:
   dcs:
     postgresql:
       use_pg_rewind: true
+      remove_data_directory_on_rewind_failure: true
+      remove_data_directory_on_diverged_timelines: true
```
WRFitch, line 6: Are we sure this is necessary, and won't happen by accident? Surely if rewind fails, we should make it possible for a user to go into the database, fix whatever's blocking rewind, then try again?

marceloneppel: About `remove_data_directory_on_rewind_failure`, if pg_rewind fail to synchronize the data from the new primary to the former primary for some reason (like pg_rewind not being able to write some file, or other things) it probably will leave the data directory in an unrecoverable state and the recommended way to solve that is to remove the data directory, according the PostgreSQL documentation (warning section from https://www.postgresql.org/docs/current/app-pgrewind.html).

The other setting, `remove_data_directory_on_diverged_timelines`, will delete the data directory only if, for some reason, it is not possible to use pg_rewind. And this process would be started in the same situation as pg_rewind would.

Both settings will not be used to delete the data directory if the connection is not possible to the former primary or if the new primary has not run a checkpoint (written data from the Write Ahead Logs files, to the final data files - something that is important to make it a consistent instance to follow). It would happen only in the situations described above.

WRFitch: ok, thanks for the detailed explanation :)

WRFitch (approved):

tests/integration/test_tls.py
```diff
@@ -65,6 +69,41 @@ async def test_mattermost_db(ops_test: OpsTest) -> None:
...
+        # Check that the primary changed.
+        assert await primary_changed(ops_test, primary), "primary not changed"
+
+        # Restart the initial primary and check the logs to ensure TLS is being used by pg_rewind.
+        await run_command_on_unit(ops_test, primary, "/charm/bin/pebble start postgresql")
+        logs = await run_command_on_unit(ops_test, replica, "/charm/bin/pebble logs")
+        assert (
+            "connection authorized: user=rewind database=postgres"
+            " SSL enabled (protocol=TLSv1.3, cipher=TLS_AES_256_GCM_SHA384, bits=256)" in logs
+        ), "TLS is not being used on pg_rewind connections"
```
MiaAltieri, line 105: Wow so thorough 🤩

MiaAltieri (approved):
Nice work! Thorough tests!

PR #46, opened 2022-09-22
Add check for primary endpoint ready

marceloneppel:
# Issue
* When relating the PostgreSQL k8s charm with other charm before the PostgreSQL charm is fully initialised (in an Active state), we can face some errors related to database connection problems. This can be checked by testing the relation between this charm and the PgBouncer k8s charm.
* Those problems happen due the PostgreSQL charm primary k8s endpoint/service is not yet correctly pointing to the primary pod (it takes some time to work).
* Jira issue: [DPE-701](https://warthogs.atlassian.net/browse/DPE-701)

# Solution
* Avoid setting the `cluster_initialised` flag (that is checked in multiple charm events) before the primary endpoint/service is fully ready by checking its accessibility.

# Context

* Files to `FOCUS` when reviewing this PR:

  * `src/charm.py`: has the call to the check, an update to the upgrade charm event handler to avoid problems withe the primary endpoint being deleted when the pod is killed and also an update to the update status event holder to avoid logging error messages when the PostgreSQL/Patroni pebble service is not ready yet.

  * `src/patroni.py`: has the implementation of the primary endpoint/service readiness check. 

  * `src/resources.yaml` has an additional port that is used to contact the primary endpoint during the initial check on charm.py pebble ready event.

  * `tests/integration/new_relations/test_new_relations.py` has an update to merge two tests in one (in a way that makes the PostgreSQL charm to be related faster to the application charm; this avoids any kind of regression on this new implementation).

* Extra details about the other files:

  * `tests/unit/test_charm.py` and `tests/unit/test_patroni.py` were updated just to match the new implementations.

# Testing
* Unit tests were updated.
* Integration tests for the relation were changed to (the same will work for legacy relations, but the charms that are used on those tests work differently, so they don't trigger the specific issue that is being solved through this PR).
* The changes from this PR were manually tested by deploying this charm, the PgBouncer k8s charm and immediately relating them (which triggers the problem before these changes).

# Release Notes
* Add check for started/ready primary endpoint/service.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -137,7 +137,7 @@ class PostgresqlOperatorCharm(CharmBase):
     def postgresql(self) -> PostgreSQL:
         """Returns an instance of the object used to interact with the database."""
         return PostgreSQL(
-            primary_host=self.primary_endpoint,
+            primary_host=self._primary_endpoint,
             current_host=self.endpoint,
             user=USER,
             password=self.get_secret("app", f"{USER}-password"),
@@ -150,7 +150,7 @@ class PostgresqlOperatorCharm(CharmBase):
         return f'{self._unit.replace("/", "-")}.{self._build_service_name("endpoints")}'
 
     @property
-    def primary_endpoint(self) -> str:
+    def _primary_endpoint(self) -> str:
         """Returns the endpoint of the primary instance's service."""
         return self._build_service_name("primary")
 
@@ -616,7 +616,7 @@ class PostgresqlOperatorCharm(CharmBase):
         return Patroni(
             self._endpoint,
             self._endpoints,
-            self.primary_endpoint,
+            self._primary_endpoint,
             self._namespace,
             self._storage_path,
             self.get_secret("app", USER_PASSWORD_KEY),
--- a/src/charm.py
+++ b/src/charm.py
@@ -137,7 +137,7 @@ class PostgresqlOperatorCharm(CharmBase):
     def postgresql(self) -> PostgreSQL:
         """Returns an instance of the object used to interact with the database."""
         return PostgreSQL(
-            primary_host=self._primary_endpoint,
+            primary_host=self.primary_endpoint,
             current_host=self.endpoint,
             user=USER,
             password=self.get_secret("app", f"{USER}-password"),
@@ -150,7 +150,7 @@ class PostgresqlOperatorCharm(CharmBase):
         return f'{self._unit.replace("/", "-")}.{self._build_service_name("endpoints")}'
 
     @property
-    def _primary_endpoint(self) -> str:
+    def primary_endpoint(self) -> str:
         """Returns the endpoint of the primary instance's service."""
         return self._build_service_name("primary")
 
@@ -616,7 +616,7 @@ class PostgresqlOperatorCharm(CharmBase):
         return Patroni(
             self._endpoint,
             self._endpoints,
-            self._primary_endpoint,
+            self.primary_endpoint,
             self._namespace,
             self._storage_path,
             self.get_secret("app", USER_PASSWORD_KEY),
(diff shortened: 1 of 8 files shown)


WRFitch (dismissed):

src/patroni.py
```diff
@@ -109,6 +110,25 @@ def are_all_members_ready(self) -> bool:
...
+            for attempt in Retrying(stop=stop_after_delay(10), wait=wait_fixed(3)):
+                with attempt:
+                    r = requests.get(
+                        f"{'https' if self._tls_enabled else 'http'}://{self._primary_endpoint}:8008/health",
+                        verify=self._verify,
+                    )
+        except RetryError:
+            return False
+
+        return r.json()["state"] == "running"
```
MiaAltieri, line 130: If the state isn't running is it worth retrying? ie
```
        try:
            for attempt in Retrying(stop=stop_after_delay(10), wait=wait_fixed(3)):
                with attempt:
                    r = requests.get(
                        f"{'https' if self._tls_enabled else 'http'}://{self._primary_endpoint}:8008/health",
                        verify=self._verify,
                    )
                    if not r.json()["state"] == "running":
                         raise endpointNotReadyError
        except RetryError: 
            return False

        return True
```

tests/integration/new_relations/test_new_relations.py
```diff
@@ -66,17 +68,11 @@ async def test_deploy_charms(ops_test: OpsTest, application_charm, database_char
...
-    # Relate the charms and wait for them exchanging some connection data.
-    await ops_test.model.add_relation(
-        f"{APPLICATION_APP_NAME}:{FIRST_DATABASE_RELATION_NAME}", DATABASE_APP_NAME
-    )
-    await ops_test.model.wait_for_idle(apps=APP_NAMES, status="active")
+        # Relate the charms and wait for them exchanging some connection data.
+        await ops_test.model.add_relation(
+            f"{APPLICATION_APP_NAME}:{FIRST_DATABASE_RELATION_NAME}", DATABASE_APP_NAME
+        )
+        await ops_test.model.wait_for_idle(apps=APP_NAMES, status="active", raise_on_blocked=True)
```
MiaAltieri, line 75: cleverly tested :)

MiaAltieri (commented):
Looks awesome!! One minor suggestion

src/charm.py
```diff
@@ -594,8 +616,8 @@ def _patroni(self):
         return Patroni(
             self._endpoint,
             self._endpoints,
+            self.primary_endpoint,
```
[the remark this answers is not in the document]
marceloneppel: Yes, I am changing that. Thanks Will!

src/patroni.py
marceloneppel: Yes. It is a great improvement! I will implement it. Thanks Mia!

WRFitch (approved):

src/charm.py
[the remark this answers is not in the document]
marceloneppel: I added it, but then I realised that it is used in the new relation file, so I reverted the change.

src/patroni.py
marceloneppel: Updated on https://github.com/canonical/postgresql-k8s-operator/pull/46/commits/fbb43329a331d55cc2aee2f304d2f5adf8bec838.

marceloneppel:
Added an additional test for relation broken on `tests/integration/new_relations/test_new_relations.py` via https://github.com/canonical/postgresql-k8s-operator/pull/46/commits/3b61943fc4e479d9866bb3cc012b005216a4cd69. The updated library code was ok on the VM charm but it was missing here, so I updated here and added the test. Now the relations are more robust.

MiaAltieri (approved):
Looks awesome, really good work here :)

PR #49, opened 2022-10-04
Add freeze db process test

marceloneppel:
# Issue
* Jira issues: [DPE-539](https://warthogs.atlassian.net/browse/DPE-539)
* The PostgreSQL charm should self heal the workload when freezing the DB process (and later unfreezing it).

# Solution
* Add a test that freezes the DB processe, unfreezing it after some time and later check that the workload recovers itself from that situation.
* Logic copied from https://github.com/canonical/postgresql-operator/pull/39.

# Context
* This test can also be run on an existing cluster like in the previous implemented test from https://github.com/canonical/postgresql-k8s-operator/pull/59.
* An mechanism to get the other instances REST API URL was created on `src/cluster.py` to enable the charm to find the primary (that is used to update the relation application databag with the primary and the replicas list) when the unit that had the process frozen is the leader.

# Testing
* The test was added on `tests/integration/ha_tests/test_self_healing.py`.

# Release Notes
* Add freeze db process test.


src/patroni.py
```diff
@@ -69,6 +70,20 @@ def _patroni_url(self) -> str:
...
 
+    def _get_alternative_patroni_url(self, attempt: AttemptManager) -> str:
+        """Get an alternative REST API URL from another member each time.
+
+        When the Patroni process is not running in the current unit it's needed
+        to use a URL from another cluster member REST API to do some operations.
+        """
+        if attempt.retry_state.attempt_number > 1:
+            url = self._patroni_url.replace(
+                self._endpoint, list(self._endpoints)[attempt.retry_state.attempt_number - 2]
```
MiaAltieri, line 81: neat 😉

tests/integration/ha_tests/test_self_healing.py
```diff
@@ -47,6 +47,63 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
...
+
+        # Un-freeze the old primary.
+        await send_signal_to_process(ops_test, primary_name, process, "SIGCONT")
+
+        # Verify that the database service got restarted and is ready in the old primary.
+        assert await postgresql_ready(ops_test, primary_name)
+
+    # Verify that a new primary gets elected (ie old primary is secondary).
+    new_primary_name = await get_primary(ops_test, app, down_unit=primary_name)
+    assert new_primary_name != primary_name
```
MiaAltieri, line 79: move this to before "Un-freeze old primary"

MiaAltieri (changes requested):
Looks AWESOME! Only one comment :)

tests/integration/ha_tests/test_self_healing.py
marceloneppel: Thanks! I'll do it.

marceloneppel: Updated on https://github.com/canonical/postgresql-k8s-operator/pull/49/commits/c07b3fc338114ff75299f9a4fc2d0efaffd9c70f.

taurus-forever (approved):
Superb!

MiaAltieri (approved):
AWESOME work here Marcelo! Happy holidays

PR #60, opened 2022-11-04
Add SIMULATE_CAN_CONNECT to the unit tests

dragomirp:
# Issue

* Jira issues: [DPE-237](https://warthogs.atlassian.net/browse/DPE-237)
* The PostgreSQL charm unit tests should use the OPS harness connectivity simulation for K8S.

# Solution

* Enable the SIMULATE_CAN_CONNECT flag during unit test initialization, tweak existing tests, add additional tests for connectivity.

# Context

* Utilize the OPS harness connectivity simulation to test for connectivity issues with the workload container.

# Testing

* The flag was enabled on tests that use ops.testing and some tests were added to cover connection checks.

# Release Notes

* Enable SIMULATE_CAN_CONNECT in unit tests.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -13,4 +13,5 @@ parts:
       - libssl-dev
       - rustc
       - cargo
-    charm-binary-python-packages: [psycopg2-binary==2.9.3, setuptools]
+    charm-binary-python-packages: [psycopg2-binary==2.9.3]
+    charm-python-packages: [setuptools]
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -13,5 +13,4 @@ parts:
       - libssl-dev
       - rustc
       - cargo
-    charm-binary-python-packages: [psycopg2-binary==2.9.3]
-    charm-python-packages: [setuptools]
+    charm-binary-python-packages: [psycopg2-binary==2.9.3, setuptools]
--- a/src/charm.py
+++ b/src/charm.py
@@ -370,7 +370,10 @@ class PostgresqlOperatorCharm(CharmBase):
         # Otherwise, each unit will create a different cluster and
         # any update in the members list on the units won't have effect
         # on fixing that.
-        if not self.unit.is_leader() and "cluster_initialised" not in self._peers.data[self.app]:
+        # Also defers when the workload is not reachable.
+        if (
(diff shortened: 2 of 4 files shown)


src/charm.py
```diff
@@ -370,7 +370,10 @@ def _on_postgresql_pebble_ready(self, event: WorkloadEvent) -> None:
         # Otherwise, each unit will create a different cluster and
         # any update in the members list on the units won't have effect
         # on fixing that.
-        if not self.unit.is_leader() and "cluster_initialised" not in self._peers.data[self.app]:
+        # Also defers when the workload is not reachable.
+        if (
+            not self.unit.is_leader() and "cluster_initialised" not in self._peers.data[self.app]
+        ) or not container.can_connect():
```
WRFitch, line 376: I believe the container must be available for `pebble_ready` to be emitted, so I'm not sure this check is necessary

WRFitch (approved):
Nice!

src/charm.py
dragomirp: The harness threw an connection error while testing. Not sure if it is misbehaviour or a possible IRL issue, so added the check just in case.

WRFitch (dismissed):

src/charm.py
marceloneppel: In which specific test the error happens? I believe it should work the way Will commented. Otherwise we can file a bug on the operator framework repo.

marceloneppel (dismissed):

src/charm.py
dragomirp: You are correct. I rechecked and the ConnectionError was thrown by harness.get_container_pebble_plan, not by the charm so there is no need to check there.

charmcraft.yaml
```diff
@@ -13,4 +13,5 @@ parts:
       - libssl-dev
       - rustc
       - cargo
-    charm-binary-python-packages: [psycopg2-binary==2.9.3, setuptools]
+    charm-binary-python-packages: [psycopg2-binary==2.9.3]
+    charm-python-packages: [setuptools]
```
WRFitch, line 17: Is this package still necessary?

WRFitch (approved):

marceloneppel (approved):

charmcraft.yaml
dragomirp: jinja2 install fails otherwise.

PR #73, opened 2023-01-09
DPE 1034 Preventing extensions enabled through relations

dragomirp:
# Issue
* [DPE-1034](https://warthogs.atlassian.net/browse/DPE-1034)
* Relating with Indico using the legacy db interface hangs

# Solution
* Changed where the charm was looking for if extensions were enabled

# Context
* The Postgresql charm should block in case relation extensions are enabled, but it was looking for them in the relation databags, not in the event data
* The Indico charm was likely waiting for requested extensions to be enabled, so it kept on waiting for the database relation even after the Postgresql charm believed to be done
* Added pytest-asyncio to suppress unit test asyncio_mode warning

# Testing

# Release Notes
Fixes bug in preventing relation extensions

[DPE-1034]: https://warthogs.atlassian.net/browse/DPE-1034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -499,6 +499,11 @@ class PostgresqlOperatorCharm(CharmBase):
                 self.unit.status = BlockedStatus(f"failed to create services {e}")
                 return
 
+    @property
+    def _has_blocked_status(self) -> bool:
+        """Returns whether the unit is in a blocked state."""
+        return isinstance(self.unit.status, BlockedStatus)
+
     def _on_get_password(self, event: ActionEvent) -> None:
         """Returns the password for a user as an action response.
 
@@ -617,11 +622,17 @@ class PostgresqlOperatorCharm(CharmBase):
         """Display an active status message if the current unit is the primary."""
         container = self.unit.get_container("postgresql")
         if not container.can_connect():
+            logger.debug("on_update_status early exit: Cannot connect to container")
+            return
+
+        if self._has_blocked_status:
+            logger.debug("on_update_status early exit: Unit is in Blocked status")
             return
 
         services = container.pebble.get_services(names=[self._postgresql_service])
         if len(services) == 0:
             # Service has not been added nor started yet, so don't try to check Patroni API.
+            logger.debug("on_update_status early exit: Service has not been added nor started yet")
             return
 
         try:
--- a/src/relations/db.py
+++ b/src/relations/db.py
@@ -91,7 +91,9 @@ class DbProvides(Object):
         application_relation_databag = event.relation.data[self.charm.app]
 
         # Do not allow apps requesting extensions to be installed.
-        if "extensions" in unit_relation_databag or "extensions" in application_relation_databag:
+        if "extensions" in event.relation.data[
+            event.app
+        ] or "extensions" in event.relation.data.get(event.unit, {}):
             logger.error(
                 "ERROR - `extensions` cannot be requested through relations"
                 " - they should be installed through a database charm config in the future"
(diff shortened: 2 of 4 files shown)


src/relations/db.py
```diff
@@ -91,7 +91,9 @@ def _on_relation_changed(self, event: RelationChangedEvent) -> None:
         application_relation_databag = event.relation.data[self.charm.app]
 
         # Do not allow apps requesting extensions to be installed.
-        if "extensions" in unit_relation_databag or "extensions" in application_relation_databag:
+        if "extensions" in event.relation.data[
```
dragomirp, line 94: Just in case, since the VM and the legacy charms do a similar check.

src/relations/db.py
```diff
@@ -101,7 +103,9 @@ def _on_relation_changed(self, event: RelationChangedEvent) -> None:
 
         # Sometimes a relation changed event is triggered,
         # and it doesn't have a database name in it.
-        database = event.relation.data[event.app].get("database")
+        database = event.relation.data[event.app].get(
+            "database", event.relation.data.get(event.unit, {}).get("database")
+        )
```
dragomirp, line 108: Just in case, since the VM and the legacy charms do a similar check.

src/relations/db.py
```diff
@@ -91,7 +91,9 @@ def _on_relation_changed(self, event: RelationChangedEvent) -> None:
         application_relation_databag = event.relation.data[self.charm.app]
 
         # Do not allow apps requesting extensions to be installed.
-        if "extensions" in unit_relation_databag or "extensions" in application_relation_databag:
+        if "extensions" in event.relation.data[
+            event.app
+        ] or "extensions" in event.relation.data.get(event.unit, {}):
```
dragomirp, line 96: The extensions is passed on with the database in the event relation data. We should also change the VM charm once this PR get accepted, since it works in the same way.

marceloneppel: Good catch @dragomirp!

taurus-forever (approved):
Nice! Tnx!

tests/integration/test_db.py
```diff
@@ -17,6 +19,33 @@
...
+        num_units=num_units,
+    ),
+    # Wait until the PostgreSQL charm is successfully deployed.
+    await ops_test.model.wait_for_idle(
+        apps=[DATABASE_APP_NAME],
+        status="active",
+        raise_on_blocked=True,
+        timeout=1000,
+        wait_for_exact_units=num_units,
+    )
```
dragomirp, line 47: Splitting the build out, so we don't have to build for each test.

src/charm.py
```diff
@@ -617,11 +622,17 @@ def _on_update_status(self, _) -> None:
         """Display an active status message if the current unit is the primary."""
         container = self.unit.get_container("postgresql")
         if not container.can_connect():
+            logger.debug("on_update_status early exit: Cannot connect to container")
+            return
+
+        if self._has_blocked_status:
+            logger.debug("on_update_status early exit: Unit is in Blocked status")
```
dragomirp, line 629: Otherwise the status of the primary will be overwritten.

tests/integration/test_db.py
```diff
@@ -95,3 +107,46 @@ async def test_finos_waltz_db(ops_test: OpsTest) -> None:
 
         # Remove the PostgreSQL application.
         await ops_test.model.remove_application(DATABASE_APP_NAME, block_until_done=True)
+
+
+@pytest.mark.db_relation_tests
+async def test_indico_db_blocked(ops_test: OpsTest) -> None:
+    """Tests if deploying and relating to Indico charm will block due to requested relations."""
```
marceloneppel, line 114: One minor fix:
```suggestion
    """Tests if deploying and relating to Indico charm will block due to requested extensions."""
```

marceloneppel (dismissed):
Thanks Dragomir! That's amazing.

taurus-forever (approved):
Cool! Tnx!

PR #74, opened 2023-01-09
Add logic to handle pgdata on cloud volumes

marceloneppel:
# Issue
Jira issue: [DPE-707](https://warthogs.atlassian.net/browse/DPE-707)

When deploying the charm to GKE or EKS Patroni cannot bootstrap the cluster (on a subdirectory of the mounted storage) correctly due to some different permissions in the mounted storage (`/var/lib/postgresql/data`).

# Solution
The needed data directory (`/var/lib/postgresql/data/pgdata`) was created by the charm on `src/charm.py` as it's running as root, and the ownership is changed to the `postgres` user, so Patroni can use it to bootstrap the cluster.

# Context
You can ignore `lib/charms/observability_libs/v1/kubernetes_service_patch.py` when reviewing this PR. It's an imported file and is the one that causes the big diff.

The `build_and_deploy` helper function from `tests/integration/test_db.py` was moved to the helpers file to be also used on `tests/integration/test_charm.py`.

Also, the Patroni and PostgreSQL ports were exposed on `src/charm.py` to enable those ports being accessible on the cloud k8s environments and with that make it possible to the units communicate with each other and check the Patroni REST API when adding multiple units (otherwise the new units are not added to the cluster).

# Testing
A new integration test was added (`test_storage_with_more_restrictive_permissions`). It simulates that the mounted storage has different permissions (equal to the ones that GKE and EKS provides in the mounted storage) and the charm can still bootstrap correctly.

# Release Notes
* Add creation of `pgdata` when the charm is deployed to enable deployments on cloud k8s environments (like GKE and EKS).
* Patch k8s service to enable PostgreSQL and Patroni ports on cloud k8s environments (like GKE and EKS).

[DPE-707]: https://warthogs.atlassian.net/browse/DPE-707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -94,9 +94,9 @@ class PostgresqlOperatorCharm(CharmBase):
             charm=self, relation="restart", callback=self._restart
         )
 
-        port1 = ServicePort(5432, name=f"{self.app.name}")
-        port2 = ServicePort(8008, name=f"{self.app.name}")
-        self.service_patcher = KubernetesServicePatch(self, [port1, port2])
+        postgresql_db_port = ServicePort(5432, name=f"{self.app.name}")
+        patroni_api_port = ServicePort(8008, name=f"{self.app.name}")
+        self.service_patcher = KubernetesServicePatch(self, [postgresql_db_port, patroni_api_port])
 
     @property
     def app_peer_data(self) -> Dict:
--- a/src/charm.py
+++ b/src/charm.py
@@ -518,6 +518,11 @@ class PostgresqlOperatorCharm(CharmBase):
                 self.unit.status = BlockedStatus(f"failed to create services {e}")
                 return
 
+    @property
+    def _has_blocked_status(self) -> bool:
+        """Returns whether the unit is in a blocked state."""
+        return isinstance(self.unit.status, BlockedStatus)
+
     def _on_get_password(self, event: ActionEvent) -> None:
         """Returns the password for a user as an action response.
 
@@ -636,11 +641,17 @@ class PostgresqlOperatorCharm(CharmBase):
         """Display an active status message if the current unit is the primary."""
         container = self.unit.get_container("postgresql")
         if not container.can_connect():
+            logger.debug("on_update_status early exit: Cannot connect to container")
+            return
+
+        if self._has_blocked_status:
+            logger.debug("on_update_status early exit: Unit is in Blocked status")
             return
 
         services = container.pebble.get_services(names=[self._postgresql_service])
         if len(services) == 0:
             # Service has not been added nor started yet, so don't try to check Patroni API.
+            logger.debug("on_update_status early exit: Service has not been added nor started yet")
             return
 
         try:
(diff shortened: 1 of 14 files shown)


src/charm.py
```diff
@@ -363,12 +369,25 @@ def _on_leader_elected(self, event: LeaderElectedEvent) -> None:
         except RetryError:
             pass  # This error can happen in the first leader election, as Patroni is not running yet.
 
+    def _create_pgdata(self, container: Container):
+        """Create the PostgreSQL data directory."""
+        path = f"{self._storage_path}/pgdata"
+        if not container.exists(path):
+            container.make_dir(
+                path, permissions=0o700, user=WORKLOAD_OS_USER, group=WORKLOAD_OS_GROUP
```
delgod, line 377: I think we should set 770. It will reduce the number of issues on OpenShift
```suggestion
                path, permissions=0o770, user=WORKLOAD_OS_USER, group=WORKLOAD_OS_GROUP
```

delgod (changes requested):

tests/integration/helpers.py
```diff
@@ -28,6 +28,35 @@
...
 
+charm = None
+
+
+async def build_and_deploy(
+    ops_test: OpsTest, num_units: int, app_name: str = DATABASE_APP_NAME, status: str = "active"
+) -> None:
+    """Builds the charm and deploys a specified number of units."""
+    global charm
+    if not charm:
```
WRFitch, line 39: I can't see anywhere else this global is used

WRFitch (approved):
A couple of small nits, but otherwise looks great!

tests/integration/helpers.py
marceloneppel: It's been used to share the already built charm between multiple tests on `tests/integration/test_charm.py` and also on `tests/integration/test_db.py`.

src/charm.py
```diff
@@ -92,6 +94,10 @@ def __init__(self, *args):
             charm=self, relation="restart", callback=self._restart
         )
 
+        port1 = ServicePort(5432, name=f"{self.app.name}")
+        port2 = ServicePort(8008, name=f"{self.app.name}")
```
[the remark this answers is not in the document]
marceloneppel: Yeah, it seems a lot better. Updated on https://github.com/canonical/postgresql-k8s-operator/pull/74/commits/3ab8351cd6b55bb052360e44de0341a162c62ad0.

tests/integration/helpers.py
taurus-forever: TIL. It looks it is a common story for us:

canonical/mysql-k8s-operator/tests/integration/high_availability/high_availability_helpers.py
147:    global mysql_charm
193:    global application_charm

taurus-forever (approved):
Tnx!

WRFitch (approved):

delgod (approved):

PR #75, opened 2023-01-11
Fix delete user call when user was not created before

marceloneppel:
# Issue
Jira issue: [DPE-991](https://warthogs.atlassian.net/browse/DPE-991)

When a charm relates to the database interface without providing a database name and later the relation is removed the PostgreSQL charm tries to delete the database user that was not created (because the database user is only created if the relation is complete: the other application needs to provide a database name for that).

# Solution
Check whether the database user exists before the charm tries to delete it.

# Context
The solution was implemented in the charm library, so it can be imported on the VM charm (the charm that was tested and initially faced the issue) on https://github.com/canonical/postgresql-operator/pull/51.

As the issue also happens on this charm, it's worth to have the fix on the charm library.

# Testing
A test simulating a relation without providing a database name was implemented.

For that, the tester application got a new relation (because it was easy to implemented on that and the data integrator - the charm used in the test that lead to the bug report - is not available yet on Charmhub).

# Release Notes
Check if a database user exists before trying to delete it.

[DPE-991]: https://warthogs.atlassian.net/browse/DPE-991?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/charms/postgresql_k8s/v0/postgresql.py
+++ b/lib/charms/postgresql_k8s/v0/postgresql.py
@@ -32,7 +32,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 6
+LIBPATCH = 7
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


dragomirp (approved):

taurus-forever (approved):
Nice!

PR #84, opened 2023-01-23
Run check-lib as a part of ci.yaml CI/CD

taurus-forever:
# Issue
We do not have a well visible warning for the outdated libraries in use
and we do not publish them automatically.

We need to be sure our libraries are up-2-date on
the new pull-request and pushed to charmhub on
merging PR and publishing the new charm revision.

# Solution
Use new `charming-action/release-libraries` and run `check-libraries` on PR (ci.yaml).

# Testing
GH Actions.

# Release Notes
Use new `charming-action/release-libraries` and run `check-libraries` on PR (ci.yaml).

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,27 +6,12 @@ on:
       - main
 
 jobs:
-  lib-check:
-    name: Check libraries
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
-      - name: Check libs
-        uses: canonical/charming-actions/check-libraries@2.1.1
-        with:
-          credentials: "${{ secrets.CHARMHUB_TOKEN }}" # FIXME: current token will expire in 2023-07-04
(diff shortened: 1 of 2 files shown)


marceloneppel (approved):

.github/workflows/release.yaml
```diff
@@ -6,27 +6,12 @@ on:
...
-      - name: Checkout
-        uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
-      - name: Check libs
-        uses: canonical/charming-actions/check-libraries@2.1.1
-        with:
-          credentials: "${{ secrets.CHARMHUB_TOKEN }}" # FIXME: current token will expire in 2023-07-04
-          github-token: "${{ secrets.GITHUB_TOKEN }}"
-
```
dragomirp, line 22: Why not keep this? It is not an expensive operation and there may be a few days between the check libs ran on the feature branch and the merging to main.

taurus-forever: Because it will be executed twice.
It is still running as a part of:
```
ci-tests:
    uses: ./.github/workflows/ci.yaml"
```

dragomirp (approved):

PR #86, opened 2023-01-27
Update clear extensions blocked status to match VM charm

dragomirp:
# Issue
Copying over changes from the VM charm to keep the logic in sync

# Solution
* Copying over changes from the VM charm to keep the logic in sync
*  Updating tls_certificates lib to latest fetched by charmcraft

# Context
* Check libs still complains, possibly due to differences in rollingops lib

# Testing

# Release Notes
* Updating tls_certificates lib
* Tweaking clearing of blocked status over relation extensions

### DIFF
--- a/lib/charms/rolling_ops/v0/rollingops.py
+++ b/lib/charms/rolling_ops/v0/rollingops.py
@@ -49,7 +49,7 @@ class SomeCharm(...):
 
 To kick off the rolling restart, emit this library's AcquireLock event. The simplest way
 to do so would be with an action, though it might make sense to acquire the lock in
-response to another event.
+response to another event. 
 
 ```python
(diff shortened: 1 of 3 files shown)


taurus-forever (approved):
LGTM. It looks like action does not replace "Libraries Out of sync" with "Libraries OK" when you update libraries... Meh... Manually set. Tnx!

dragomirp:
> LGTM. It looks like action does not replace "Libraries Out of sync" with "Libraries OK" when you update libraries... Meh... Manually set. Tnx!

Check libs is actually correct, we have a one white space difference between what is in our charm and what it pulls from upstream.

marceloneppel (approved):

taurus-forever (approved):
Nice catch!

PR #89, opened 2023-02-02
Poetry

dragomirp:
# Issue
* [DPE-1133](https://warthogs.atlassian.net/browse/DPE-1133)
* Add poetry dependency management based off of the mongo-k8s charm

# Solution
* Add poetry dependency management based off of the mongo-k8s charm

# Context
* Updated charm libs and switched to `data-interfaces`
* Updated pypi dependencies
* Includes secrets inheritance from [PR#90](https://github.com/canonical/postgresql-k8s-operator/pull/90)

# Testing

# Release Notes

[DPE-1133]: https://warthogs.atlassian.net/browse/DPE-1133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


charmcraft.yaml
```diff
@@ -13,5 +13,5 @@ parts:
       - libssl-dev
       - rustc
       - cargo
-    charm-binary-python-packages: [psycopg2-binary==2.9.3]
+    charm-binary-python-packages: [psycopg2-binary==2.9.5]
```
dragomirp, line 16: We need to figure out how to add/build binaries that work to remove this. Using other `parts` plugins seems like an option, but it will take time to figure out and we'll need to verify that such a custom build will keep on being supported.

pyproject.toml
```diff
@@ -1,6 +1,58 @@
...
+coverage = {extras = ["toml"], version = "^7.0.5"}
+pytest = "^7.1.1"
+pytest-asyncio = "^0.20.3"
+
+[tool.poetry.group.integration.dependencies]
+lightkube = "^0.12.0"
+pytest = "^7.1.1"
+pytest-operator = "^0.23.0"
+juju = "^2" # Latest juju 2
+psycopg2 = {version = "^2.9.5", extras = ["binary"]}
```
dragomirp, line 50: Corresponds to environment deps in `tox.ini`. Poetry will pull them only if requested with `--with`

tests/integration/test_db.py
```diff
@@ -89,19 +89,17 @@ async def test_indico_db_blocked(ops_test: OpsTest) -> None:
...
+            "indico",
+            channel="stable",
+            application_name="indico1",
+            num_units=APPLICATION_UNITS,
+        )
+        await ops_test.model.deploy(
+            "indico",
+            channel="stable",
+            application_name="indico2",
+            num_units=APPLICATION_UNITS,
```
dragomirp, line 102: Looks like doing them in one go can error out when ci runners are under heavy load.

tox.ini
```diff
@@ -23,134 +23,95 @@ passenv =
   CHARM_BUILD_DIR
   MODEL_SETTINGS
   PYTEST_SKIP_DEPLOY
+deps =
+    poetry
```
dragomirp, line 27: Only install poetry in all the virtual envs.

tox.ini
```diff
@@ -23,134 +23,95 @@ passenv =
...
-    pytest -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs} --durations=0
+    poetry install --with integration
+    poetry export -f requirements.txt -o requirements.txt
+    poetry run pytest -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs} --durations=0
+
+[testenv:poetry-lock]
+description = Install, lock and export poetry dependencies
+commands =
+    poetry lock
+    poetry export -f requirements.txt -o requirements.txt
```
dragomirp, line 117: Auxiliary command so that we can update the lock file from `tox` when changing `pyproject.toml`. We can do this with an external, system installed poetry as well.

dragomirp:
If we copy over the renovate bot config as well, we should be able to get PRs for updated dependencies. Not strictly necessary for switching dependency management though.

charmcraft.yaml
taurus-forever: Can SNAP magically build it for us from "psycopg2"?
Drawback: PR build time will increase.

Another option is [apt package](https://packages.ubuntu.com/lunar/python3-psycopg2) or build it from sources in fat-SNAP.

tests/integration/test_db.py
```diff
@@ -89,19 +89,17 @@ async def test_indico_db_blocked(ops_test: OpsTest) -> None:
...
         )
```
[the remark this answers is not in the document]
taurus-forever: Good hint to share with team members!

taurus-forever (approved):
Tnx! Let's try to stabilize env with Poetry!

charmcraft.yaml
dragomirp: AFAIK we build snaps for the charm's workload while here we need the binary in the charm itself, so for K8S at least I doubt it can help us. The deb package we need is `libpq5`, but I couldn't figure out a way to pull it at runtime using `charmcraft.yaml` and the `charm` parts plugin. Another option we can explore is to use a pure Python postgresql client in the charm, but either way we'll need to spend some time figuring out what the best approach would be and implement it.

marceloneppel: I tried the apt package in the past, but some errors happened when I deployed the charm. Some more investigation is needed to figure out how to make it work or to find another option, like Dragomir commented.

marceloneppel (approved):
LGTM!

PR #90, opened 2023-02-05
Pass on secrets for ci.yaml to run

dragomirp:
# Issue
* CI on the main branch fails due to ci.yaml check libs task not getting its credentials for charmhub

# Solution
* Add a directive to inherit secrets, so charmhub token is present for check libs

# Context
* Tested with a forked version of the repo, change appeared to work and the check libs task seemed to have `credentials` set

# Testing

# Release Notes

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -8,6 +8,7 @@ on:
 jobs:
   ci-tests:
     uses: ./.github/workflows/ci.yaml
+    secrets: inherit
 
   release-to-charmhub:
     name: Release to CharmHub


marceloneppel (approved):
Great! Thanks Dragomir!

taurus-forever (approved):
1) apologies, it is my regression. 
2) let's merge this for now to fix CI.
3) we need to revise a solution here... and maybe extract lib-check into separate workflow , will discuss with Carl, re: DA021 Braindump.

dragomirp:
Looks like in the mean time there was an update of `indico` and that it breaks the tests in the main branch (possibly due to old `python-juju`). This does not happen with updated dependencies, so I'll roll this change in [PR#89](https://github.com/canonical/postgresql-k8s-operator/pull/89)

PR #95, opened 2023-02-16
Cache charmcraft pack container, skip unstable tests except on schedule

carlcsaposs-canonical:
Ported from https://github.com/canonical/mysql-k8s-operator/pull/146

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -10,7 +10,7 @@ import pytest
 from pytest_operator.plugin import OpsTest
 
 
-@pytest.fixture
+@pytest.fixture(scope="module")
 def ops_test(ops_test: OpsTest) -> OpsTest:
     if os.environ.get("CI") == "true":
         # Running in GitHub Actions; skip build step


.github/workflows/ci.yaml
```diff
@@ -1,11 +1,17 @@
...
+
 on:
   pull_request:
     paths-ignore:
       - '.gitignore'
       - '.jujuignore'
       - 'LICENSE'
       - '**.md'
+  schedule:
+    - cron: '53 0 * * *' # Daily at 00:53 UTC
```
dragomirp, line 13: Not an issue with the PR itself, but we may consider running the scheduled tests during slightly different times to not stampede on the runners.

tests/integration/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+
+import json
+import os
+from pathlib import Path
+
+import pytest
+from pytest_operator.plugin import OpsTest
+
+
+@pytest.fixture(scope="module")
```
dragomirp, line 13: I think we can even go with session here, it should never change during execution.

dragomirp (approved):

.github/workflows/ci.yaml
taurus-forever: Starting all MySQL and PostgreSQL repos simultaneously at 00:53 might be noticeable in [Los Angeles](https://www.timeanddate.com/worldclock/converter.html?iso=20230217T005000&p1=1440&p2=137) (16:53 local time).

taurus-forever (approved):

PR #98, opened 2023-02-23
Hide instance on backup

marceloneppel:
# Issue
Jira issue: [DPE-1270](https://warthogs.atlassian.net/browse/DPE-1270)
The PostgreSQL unit that is performing the backup operation isn't being hidden from PgBouncer or other client applications to avoid that the connections are impacted by the backup operation.

# Solution
Temporarily remove the endpoint of the unit that is performing the backup from the k8s service by removing some pod labels (which are used in the label selector of the service).

# Context
The `_patch_pod_labels` method was just renamed to not be private anymore as it's being used outside `charm.py`.

# Testing
Tests will be implemented on [DPE-556](https://warthogs.atlassian.net/browse/DPE-556).

# Release Notes
Hide instance on backup.

[DPE-1270]: https://warthogs.atlassian.net/browse/DPE-1270?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-556]: https://warthogs.atlassian.net/browse/DPE-556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/backups.py
+++ b/src/backups.py
@@ -151,14 +151,21 @@ class PostgreSQLBackups(Object):
 
         try:
             self.charm.unit.status = MaintenanceStatus("creating backup")
-            self.charm.patch_pod_labels(self.charm.unit.name, add_labels=False)
+
+            # Remove the pod labels to hide this unit from the replicas K8S service.
+            if self.charm.app.planned_units() > 1:
+                self.charm.patch_pod_labels(self.charm.unit.name, add_labels=False)
+
             self._execute_command(
                 ["pgbackrest", f"--stanza={self.charm.cluster_name}", "--type=full", "backup"]
             )
             event.set_results({"backup-status": "backup created"})
         except pebble.ExecError as e:
             event.fail(f"Failed to backup PostgreSQL with error: {str(e)}")
+
+        # Add the labels back to the pod.
         self.charm.patch_pod_labels(self.charm.unit.name)
+
         self.charm.unit.status = ActiveStatus()
 
     def _on_list_backups_action(self, event) -> None:
--- a/src/backups.py
+++ b/src/backups.py
@@ -151,12 +151,14 @@ class PostgreSQLBackups(Object):
 
         try:
             self.charm.unit.status = MaintenanceStatus("creating backup")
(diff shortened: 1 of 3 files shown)


dragomirp (approved):

src/backups.py
```diff
@@ -151,12 +151,21 @@ def _on_create_backup_action(self, event) -> None:
 
         try:
             self.charm.unit.status = MaintenanceStatus("creating backup")
+
+            # Remove the pod labels to hide this unit from the replicas K8S service.
+            if self.charm.app.planned_units() > 1:
+                self.charm.patch_pod_labels(self.charm.unit.name, add_labels=False)
```
delgod, line 157: maybe additionally we need to run `update pg_database set datallowconn = false where datname = 'databasename';` to hide db instance from pgbouncer

delgod (changes requested):

src/backups.py
marceloneppel: After testing it, I saw that it only takes effect after connections are closed and new connections are created on PgBouncer side. So it doesn’t block new connections from the client applications immediately. Also, it blocks connections on all instances because the database is the same for reads and writes on all instances.

marceloneppel:
Closing the PR as PgBouncer is caching the IPs behind the PostgreSQL replicas K8S service (so PgBouncer continues to send queries to the instance that is performing the backup).

Also because updating the database to deny connections (with `datallowconn = false`) only takes effect after some time, when connections are closed and new connections are created on PgBouncer side, and it blocks connections on all instances (including the primary), not only in the replica that is doing the backup (because PgBouncer connects to each database that it has created for the client application, either when it’s a read/write or read-only connection - in those databases that it authenticates the client application user and later run the SQL statements, so the database is the same on all the DB instances).

I’ll rework the PR using another approach (probably using pg_hba.conf, but need to test it first). I’ll open another PR with the new code.

PR #105, opened 2023-03-07
All dep groups set to optional

dragomirp:
# Issue
* Poetry groups get installed in tox environments that don't need them

# Solution
* Mark all dependency groups as optional, so they get installed only when pulled explicitly

# Context
* Also fixed some typos and bumped dependency versions

# Testing

# Release Notes
Housekeeping


pyproject.toml
```diff
@@ -2,7 +2,7 @@
 # See LICENSE file for licensing details.
 
 [tool.poetry]
-name = "postgesql-k8s"
+name = "postgresql-k8s"
```
taurus-forever, line 5: o_0 how it worked?

taurus-forever (approved):

pyproject.toml
dragomirp: This is just poetry metadata, it requires it but it's not really used for anything.

dragomirp:
New libjuju is out (2.9.42.1). Updating as part of this PR.

taurus-forever:
> 2.9.42.1

The [changelog](https://github.com/juju/python-libjuju/compare/2.9.38...2.9.42.1) says: no changes expected.

marceloneppel (approved):

PR #135, opened 2023-04-10
Test network cut

marceloneppel:
## Issue
Jira ticket: [DPE-540](https://warthogs.atlassian.net/browse/DPE-540)
PostgreSQL should self heal after cutting the network connection and reestablishing it.

## Solution
* Add a test that uses Chaos Mesh to limit the network traffic to a PostgreSQL unit and later restore the network traffic also checking that no writes (done by a continuous writes application) were missed.

In `tests/integration/ha_tests/application-charm/src/continuous_writes.py`:
* Changed the continuous writes logic to have a timeout that ends the current connection and creates a new one to continue the writes (this was needed because sometimes when the PostgreSQL DB process is frozen - in freeze DB process test - the application hasn't closed the connection until the DB process was unfrozen, preventing new writes to the new primary that was elected).

In `tests/integration/ha_tests/helpers.py`:
* Added `check_cluster_is_updated ` helper function to avoid duplicate code in the tests.
* Changed pgkill to use `-x` instead of `-f` (which enables the test to use SIGTERM instead of SIGINT).

[DPE-540]: https://warthogs.atlassian.net/browse/DPE-540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


taurus-forever (approved):
LGTM, but I would simplify shell scripts using example from https://github.com/canonical/mysql-k8s-operator/pull/131/files

delgod (approved):

PR #146, opened 2023-04-18
Update conf only when the container is connectable

dragomirp:
## Issue
[Sometimes](https://github.com/canonical/pgbouncer-k8s-operator/actions/runs/4719644966/jobs/8393434027) there is an unhandled exception while regenerating the patroni configuration due to the charm being unable to read the postgresql version from the workload container

## Solution
Check if the workload container is connectable before regenerating the config

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -237,7 +237,6 @@ class PostgresqlOperatorCharm(CharmBase):
             logger.debug(
                 "Early exit on_peer_relation_changed: Waiting for container to become available"
             )
-            event.defer()
             return
 
         # Validate the status of the member before setting an ActiveStatus.
--- a/src/charm.py
+++ b/src/charm.py
@@ -238,6 +238,7 @@ class PostgresqlOperatorCharm(CharmBase):
                 "Early exit on_peer_relation_changed: Waiting for container to become available"
             )
             return
+        self.update_config()
 
         # Validate the status of the member before setting an ActiveStatus.
         if not self._patroni.member_started:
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -232,6 +232,13 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent) -> None:
 
         # Update the list of the cluster members in the replicas to make them know each other.
         # Update the cluster members in this unit (updating patroni configuration).
+        container = self.unit.get_container("postgresql")
+        if not container.can_connect():
+            logger.debug(
+                "Deferring on_peer_relation_changed: Waiting for container to become available"
+            )
+            event.defer()
+            return
```
dragomirp, line 241: I'm not sure if we should defer here. If the workload is not fully up, the configuration should be generated once pebble is ready. Thoughts?

tests/unit/test_patroni.py
```diff
@@ -166,8 +166,10 @@ def test_render_patroni_yml_file(self, _render_file, _rock_postgresql_version):
         )
         self.assertIn("ssl_key_file: /var/lib/postgresql/data/key.pem", expected_content_with_tls)
 
+    @patch("patroni.stop_after_delay", return_value=stop_after_delay(0))
+    @patch("patroni.wait_fixed", return_value=wait_fixed(0))
```
dragomirp, line 170: So we don't wait for tenacity.

src/charm.py
marceloneppel: I think it's ok to just return here, and leave to pebble ready event to update the config. If, later, the config needs to be updated again, this or some other hooks will be fired again (to call update_config).

taurus-forever (approved):

taurus-forever:
Nice trace! Tnx!

PR #154, opened 2023-05-04
Fix resources creation

marceloneppel:
## Issue
Jira ticket: [DPE-1590](https://warthogs.atlassian.net/browse/DPE-1590)

The charm cannot be deployed in older microk8s (like 1.22 - Kubeflow team tested it - and 1.21 - Commercial Systems team tested it on PS5).

## Solution
Don't create resources that are already being created by Patroni (like the role and role binding). Only create the custom services (used mainly in relations) in the charm code.

Don't replace resources when they already exist (in fact, for the custom services, it's not needed to update them, as they are already created with all the needed metadata). This will be improved, if needed, on [DPE-1470](https://warthogs.atlassian.net/browse/DPE-1470) to update existing k8s resources.

The unit tests were updated to avoid regressions.

This fix will enable Commercial Systems to use our charm with the [Temporal K8S charm](https://charmhub.io/temporal-k8s).

[DPE-1590]: https://warthogs.atlassian.net/browse/DPE-1590?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-1470]: https://warthogs.atlassian.net/browse/DPE-1470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/charm.py
```diff
@@ -528,12 +528,9 @@ def _create_resources(self) -> None:
...
             # The 409 error code means that the resource was already created
-            # or has a higher version. This can happen if Patroni creates a
-            # resource that the charm is expected to create.
-            if e.status.code == 409:
-                logger.debug("replacing resource: %s.", str(resource.to_dict()))
-                client.replace(resource)
-            else:
+            # or has a higher version. This can happen when multiple calls are
+            # made to this function.
+            if e.status.code != 409:
```
delgod, line 533: In general, if we need to modify/change objects created by patroni - it is better to use the `client.apply(resource)` call.
So configuration will be merged by server-side apply.

delgod (approved):

src/charm.py
marceloneppel: Thanks for the info, @delgod! I'll use that on [DPE-1470](https://warthogs.atlassian.net/browse/DPE-1470).

taurus-forever (approved):
Nice!

PR #165, opened 2023-05-24
Add 2nd batch of backup/restore unit tests

marceloneppel:

### DIFF
--- a/tests/unit/test_backups.py
+++ b/tests/unit/test_backups.py
@@ -1,0 +1,463 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+import datetime
+import unittest
+from unittest.mock import MagicMock, PropertyMock, call, patch
+
+from botocore.exceptions import ClientError
+from ops import ActiveStatus, BlockedStatus, MaintenanceStatus
+from ops.pebble import Change, ChangeError, ChangeID, ExecError
+from ops.testing import Harness
+
+from charm import PostgresqlOperatorCharm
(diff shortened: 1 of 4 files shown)


tests/unit/test_backups.py
```diff
@@ -0,0 +1,463 @@
...
+        _can_use_s3_repository.assert_called_once()
+        _initialise_stanza.assert_called_once()
+        _start_stop_pgbackrest_service.assert_called_once()
+
+    @patch("charm.PostgreSQLBackups._change_connectivity_to_database")
+    @patch("charm.PostgreSQLBackups._list_backups")
+    @patch("charm.PostgreSQLBackups._execute_command")
+    @patch("charm.PostgresqlOperatorCharm.is_primary", new_callable=PropertyMock)
+    @patch("charm.PostgreSQLBackups._upload_content_to_s3")
+    @patch("datetime.datetime")
```
dragomirp, line 156: Does patching `charm.PostgreSQLBackups.datetime` not work?

marceloneppel: Unfortunately, it doesn't work. I also tested some other approaches, like `charm.datetime`, but none of them worked. Then I added the approach from this PR, that was copied from MySQL VM charm.

marceloneppel: Cool, let me try it here. Thanks for the suggestion Dragomir.

dragomirp (approved):

taurus-forever (approved):
Great example of excellence!

PR #164, opened 2023-05-24
[DPE-556] Add 1st batch of backup/restore unit tests

marceloneppel:
## Issue
The backup and restore features don't have unit tests.

## Solution
Add unit tests for 1/3 of the implemented methods.

The remaining 2/3 of unit tests were implemented on https://github.com/canonical/postgresql-k8s-operator/pull/165 and https://github.com/canonical/postgresql-k8s-operator/pull/166 to keep this PR small.

A fix was made on `src/backups.py` to really rollback the stanza name when the `pgbackrest check command fails`. It was incorrectly setting the stanza name again in the relation data, instead of clearing it.

PR #166, opened 2023-05-24
[DPE-556] Add 3rd batch of backup/restore unit tests

marceloneppel:
## Issue
The backup and restore features don't have unit tests.

## Solution
Add unit tests for 1/3 of the implemented methods.

The remaining 2/3 of unit tests were implemented on https://github.com/canonical/postgresql-k8s-operator/pull/164 and https://github.com/canonical/postgresql-k8s-operator/pull/165 to keep this PR small.

The logic that fixes slashes in the S3 path was moved in `src/backups.py` to handle that in the right place (after a default path is set).

### DIFF
--- a/src/backups.py
+++ b/src/backups.py
@@ -683,14 +683,14 @@ Stderr:
             )
             return {}, missing_required_parameters
 
-        # Retrieve the backup path, strip its slashes and add a "/" in the beginning of the path.
-        s3_parameters["path"] = f'/{s3_parameters["path"].strip("/")}'
-
         # Add some sensible defaults (as expected by the code) for missing optional parameters
         s3_parameters.setdefault("endpoint", "https://s3.amazonaws.com")
         s3_parameters.setdefault("region")
         s3_parameters.setdefault("path", "")
         s3_parameters.setdefault("s3-uri-style", "host")
+
+        # Retrieve the backup path, strip its slashes and add a "/" in the beginning of the path.
+        s3_parameters["path"] = f'/{s3_parameters["path"].strip("/")}'
 
         return s3_parameters, []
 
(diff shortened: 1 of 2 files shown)


src/backups.py
```diff
@@ -683,15 +683,15 @@ def _retrieve_s3_parameters(self) -> Tuple[Dict, List[str]]:
...
-
         # Add some sensible defaults (as expected by the code) for missing optional parameters
         s3_parameters.setdefault("endpoint", "https://s3.amazonaws.com")
         s3_parameters.setdefault("region")
         s3_parameters.setdefault("path", "")
         s3_parameters.setdefault("s3-uri-style", "host")
 
+        # Retrieve the backup path, strip its slashes and add a "/" in the beginning of the path.
+        s3_parameters["path"] = f'/{s3_parameters["path"].strip("/")}'
+
```
taurus-forever, line 694: Maybe consider to mimic MySQL: https://github.com/canonical/mysql-k8s-operator/blob/main/lib/charms/mysql/v0/backups.py#L142-L151

We strip trailing spaces and unnecessary slashes to make sure no garbage comes from users, e.g. endlpoint="https://s3.amazonaws.com//"  or patl="alex " might cause a lot of fun with troubleshooting.

P.S. sure, good to validate/trim/strip on s3-integrator level... but since we strip path here, maybe worse to sync with MySQL approach.

taurus-forever (approved):
One possible improvent here.

src/backups.py
marceloneppel: Great suggestion. I'm gonna update the PR to sync with MySQL approach.

marceloneppel: Updated on https://github.com/canonical/postgresql-k8s-operator/pull/166/commits/4751cbe677d5e9a39c88f0d2d431cc8c5d3f7a6c.

PR #170, opened 2023-05-30
Add plugins support

marceloneppel:

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -10,6 +10,7 @@ from typing import Dict, List, Optional
 from charms.observability_libs.v1.kubernetes_service_patch import KubernetesServicePatch
 from charms.postgresql_k8s.v0.postgresql import (
     PostgreSQL,
+    PostgreSQLEnableDisableExtensionError,
     PostgreSQLUpdateUserPasswordError,
 )
 from charms.postgresql_k8s.v0.postgresql_tls import PostgreSQLTLS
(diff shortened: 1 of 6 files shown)


src/relations/db.py
```diff
@@ -89,34 +89,57 @@ def _on_relation_changed(self, event: RelationChangedEvent) -> None:
...
+        """Set up the relation to be used by the application charm."""
+        # Do not allow apps requesting extensions to be installed
+        # (let them now about config options).
+        requested_extensions, disabled_extensions = self._get_extensions(relation)
+        if disabled_extensions:
             logger.error(
-                "ERROR - `extensions` cannot be requested through relations"
-                " - they should be installed through a database charm config in the future"
+                f"ERROR - `extensions` ({', '.join(disabled_extensions)}) cannot be requested through relations"
+                " - they should be enabled through a database charm config"
```
delgod, line 119:
```suggestion
                f"ERROR - `extensions` ({', '.join(disabled_extensions)}) cannot be requested through relations"
                " - Please enable extensions through `juju config` and add the relation again."
```

delgod (approved):

src/charm.py
```diff
@@ -266,9 +272,55 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent) -> None:
...
+                    "Early exit on_config_changed: db relation requested extensions that are still disabled"
+                )
+                return
+
+        for relation in self.model.relations.get("db-admin", []):
+            if not self.legacy_db_admin_relation.set_up_relation(relation):
+                logger.debug(
+                    "Early exit on_config_changed: db-admin relation requested extensions that are still disabled"
+                )
+                return
```
dragomirp, line 303: We can loop over the relations in one go
```suggestion
        for relation in [
            *self.model.relations.get("db", []),
            *self.model.relations.get("db-admin", []),
        ]:
            if not self.legacy_db_relation.set_up_relation(relation):
                logger.debug(
                    "Early exit on_config_changed: legacy relation requested extensions that are still disabled"
                )
                return
```

marceloneppel: You're right. I forgot that the behaviour is the same. I'm gonna update it.

PR #94, opened 2023-02-14
Upgrade to jammy

marceloneppel:
# Issue
Jira issue: https://warthogs.atlassian.net/browse/DPE-1411
The charm is still on Focal.

# Solution
Upgrade all the bases and all the deployments to use Jammy.

# Context
The Discourse charm started to request extensions (the version that requests extensions was promoted to stable today - 2023-02-28): https://github.com/canonical/discourse-k8s-operator/commit/08830d595de9bd391e74d5b0740f0390b8861a1e. So, the test was changed in this PR.

# Testing
Used the existing tests to confirm that the charm works on Jammy.

# Release Notes
Upgrade charm to Jammy.

PR #180, opened 2023-06-07
[DPE-2057] Enable Discourse-K8S test

marceloneppel:
## Issue
The Discourse K8S charm test was disabled on https://github.com/canonical/postgresql-k8s-operator/pull/94 because it started to request extensions through the relation.

## Solution
In this PR, we enable the test back again and use the [new plugins/extensions support](https://github.com/canonical/postgresql-k8s-operator/pull/170) to make it work back again.

Now, Discourse needs only the `db` relations (rather than the `db-admin` relations, that was previously used by Discourse itself to install the extensions). So, the test was moved to the `db` relation integration test file.

Some unit tests were copied from https://github.com/canonical/postgresql-operator/pull/139 to improve the coverage of this charm.

### DIFF
--- a/src/relations/db.py
+++ b/src/relations/db.py
@@ -201,6 +201,8 @@ class DbProvides(Object):
             )
             return False
 
+        self._update_unit_status(relation)
+
         return True
 
     def _check_for_blocking_relations(self, relation_id: int) -> bool:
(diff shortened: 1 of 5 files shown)


dragomirp (approved):

taurus-forever (approved):
Incredible!

PR #185, opened 2023-06-14
[DPE-1994] Patroni COS

dragomirp:
## Issue
No COS integration for Patroni

## Solution
Add COS integration for Patroni
Add explicit log file paths for PG and Patroni


src/charm.py
```diff
@@ -124,6 +124,17 @@ def __init__(self, *args):
         patroni_api_port = ServicePort(8008, name="api")
         self.service_patcher = KubernetesServicePatch(self, [postgresql_db_port, patroni_api_port])
 
+    def _generate_metrics_jobs(self, enable_tls: bool) -> Dict:
+        """Generate spec for Prometheus scraping."""
+        return [
+            {"static_configs": [{"targets": [f"*:{METRICS_PORT}"]}]},
```
dragomirp, line 130: PG exporter is always served on http

src/charm.py
```diff
@@ -124,6 +124,17 @@ def __init__(self, *args):
...
+            {
+                "static_configs": [{"targets": ["*:8008"]}],
+                "scheme": "https" if enable_tls else "http",
+                "tls_config": {"insecure_skip_verify": True},
```
dragomirp, line 134: Certificate is likely self signed

src/charm.py
```diff
@@ -124,6 +124,17 @@ def __init__(self, *args):
         patroni_api_port = ServicePort(8008, name="api")
         self.service_patcher = KubernetesServicePatch(self, [postgresql_db_port, patroni_api_port])
 
+    def _generate_metrics_jobs(self, enable_tls: bool) -> Dict:
+        """Generate spec for Prometheus scraping."""
+        return [
+            {"static_configs": [{"targets": [f"*:{METRICS_PORT}"]}]},
+            {
+                "static_configs": [{"targets": ["*:8008"]}],
+                "scheme": "https" if enable_tls else "http",
```
dragomirp, line 133: Patroni will use TLS if available, so the scheme must change conditionally.

src/charm.py
```diff
@@ -1110,17 +1119,20 @@ def update_config(self) -> None:
...
         self._patroni.reload_patroni_configuration()
-        self.unit_peer_data.update({"tls": "enabled" if enable_tls else ""})
+        self.unit_peer_data.update({"tls": "enabled" if self.is_tls_enabled else ""})
 
         # Restart PostgreSQL if TLS configuration has changed
         # (so the both old and new connections use the configuration).
         if restart_postgresql:
+            self.metrics_endpoint.update_scrape_job_spec(
+                self._generate_metrics_jobs(self.is_tls_enabled)
+            )
```
dragomirp, line 1135: If we are restarting due to TLS, we must also update the patroni endpoint scheme

src/constants.py
```diff
@@ -20,6 +20,10 @@
 WORKLOAD_OS_GROUP = "postgres"
 WORKLOAD_OS_USER = "postgres"
 METRICS_PORT = "9187"
-POSTGRES_LOG_FILES = ["/var/log/pgbackrest", "/var/log/postgresql"]
+POSTGRES_LOG_FILES = [
+    "/var/log/pgbackrest",
```
dragomirp, line 24: I'm not sure how pgbackrest calls its log files

src/grafana_dashboards/postgresql-patroni-metrics.json
```diff
@@ -0,0 +1,3461 @@
...
+      "pluginVersion": "9.2.13",
+      "targets": [
+        {
+          "datasource": {
+            "type": "datasource",
+            "uid": "${prometheusds}"
+          },
+          "editorMode": "code",
+          "exemplar": false,
+          "expr": "patroni_sync_standby{service_name=~\"$service_name\",scope=~\"$scope_name\"}",
```
dragomirp, line 1378: This will not be populated until we upgrade to Patroni 3.0.2

src/constants.py
marceloneppel: Hmm, I saw different logs related to backups, restores and stanza initialisation, but I don't know the complete list, so if the directory works here, that's fine.

marceloneppel (approved):

dragomirp:
Retested with `grafana-agent`. TLS worked, however we must still keep the `insecure_skip_verify` directive, since it seems we need to upload the CA cert to the agent charm to be able to use `ca_file` directive.

src/charm.py
taurus-forever: IMHO, we must use insecure_skip_verify for self-signed certificates only.
If CA is real -> real cert with CA known to grafana-agent => we should not set this option.
Please report it as a TODO ticket in backlog (attached to the topic of will TLS support on DB<>COS, as we need to add TLS support for *_exporters too).

src/constants.py
taurus-forever: I have feeling the folder doesn't work. But let's test it this way.

taurus-forever (approved):
Nice! Thank you!

src/charm.py
dragomirp: [DPE-2133](https://warthogs.atlassian.net/browse/DPE-2133)

PR #188, opened 2023-06-19
Internal changes to the charm shouldn't be released

dragomirp:
## Issue
Charm is published to charmhub on each push to main even if there are no changes to the charm itself

## Solution
Try to limit the amount of releases based on the content of the push

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -8,6 +8,7 @@ on:
       - main
     paths-ignore:
       - 'tests/**'
+      - 'docs/**'
       - renovate.json
       - poetry.lock
       - pyproject.toml
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,6 +6,12 @@ on:
   push:
     branches:
       - main
+    paths-ignore:
+      - 'tests/**'
+      - renovate.json
+      - poetry.lock
+      - pyproject.toml
+      - '.github/**.md'
 
 jobs:
   ci-tests:


.github/workflows/release.yaml
```diff
@@ -6,6 +6,12 @@ on:
   push:
     branches:
       - main
+    paths-ignore:
+      - 'tests/**'
+      - renovate.json
+      - poetry.lock
+      - pyproject.toml
```
dragomirp, line 13: If only package metadata in `pyproject.toml` or `poetry.lock` changes without changing `requirements.txt` we shouldn't publish.

marceloneppel (approved):

.github/workflows/release.yaml
taurus-forever: @carlcsaposs-canonical we will need to find a way to make release.yaml workflow shared via data-platform-workflows to share such changes across our repos (avoiding copy&paste). Sure, no rush here!

taurus-forever: @dragomirp should "docs" also be in this path? Or we block them on CI/CD level (skipping building).
Q: shy do we spend runners time on building if we not release? Should we [skip](https://github.com/canonical/postgresql-k8s-operator/blob/main/.github/workflows/ci.yaml#L11) building too?

taurus-forever (commented):
In general LGTM, Q about the implementation.

.github/workflows/release.yaml
dragomirp: > @dragomirp should "docs" also be in this path? Or we block them on CI/CD level (skipping building). 
Docs will still execute the workflow. If we don't intend to do anything automated with them, we can add them to the ignores.

> Q: shy do we spend runners time on building if we not release? Should we [skip](https://github.com/canonical/postgresql-k8s-operator/blob/main/.github/workflows/ci.yaml#L11) building too?
This will prevent the entire pipeline from running (both CI workflow and Release jobs). In the future, when we aren't starved for runners, we may consider running just the CI flow as a sanity check, but ATM I don't think it's worth it.

taurus-forever (approved):

PR #196, opened 2023-06-30
[DPE-1999] Fixed TLS race condition

marceloneppel:
## Issue
Sometimes one or more units became stuck in 'awaiting for cluster to start'/'awaiting for member to start' status due to race conditions between Patroni startup and TLS enablement on the workload (the charm sets the flag the changes all communications to use TLS while Patroni is already starting with TLS disabled).

## Solution
Set the `tls` flag only when Patroni is already running (and in that case reload its configuration) or when the workload is not running (checking pebble services; when it's safe to update the charm to use TLS in the communications, as Patroni will start with TLS enabled - because it will read the config with TLS enabled at the time it starts in another hook, like pebble ready).


lib/charms/postgresql_k8s/v0/postgresql_tls.py
```diff
@@ -139,7 +141,10 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
         self.charm.set_secret(SCOPE, "ca", event.ca)
 
         try:
-            self.charm.push_tls_files_to_workload()
+            if not self.charm.push_tls_files_to_workload():
```
dragomirp, line 144: We'll have to update the other charms (PG VM and PGB K8S) or check explicitly for false, so that None will not trigger a defer.

dragomirp (approved):

lib/charms/postgresql_k8s/v0/postgresql_tls.py
marceloneppel: Exactly. I'm preparing PRs for the other PG and PgBouncer repos.

taurus-forever: Are we creating a strict binding: PG rev 109+ requires PGB revision 58+?
Should we start building compatibility matrix inside charm.py to check before deployment and warn users on incompatibilities... or create backward compatible code.

taurus-forever (approved):
LGTM, one comment.

lib/charms/postgresql_k8s/v0/postgresql_tls.py
marceloneppel: No Alex, we aren't creating that strict binding. What we need to do is update PgBouncer code in a similar way as here (adding the returns) when we upgrade the TLS lib there (in [this PR](https://github.com/canonical/pgbouncer-k8s-operator/pull/111)).

Issue #192, opened 2023-06-22
Error on use of extra_user_roles="admin" with DatabaseRequires


AmberCharitos:
## Steps to reproduce

```
# clone trino charm repo:
git clone https://github.com/canonical/trino-k8s-operator.git

# checkout postgres-relation branch
git checkout postgres-relation

# pack charm
cd trino-k8s-charm
charmcraft pack

# deploy trino charm
juju deploy ./trino-k8s_ubuntu-22.04-amd64.charm --resource trino-image=trinodb/trino:418

# deploy postgres charm
juju deploy postgresql-k8s --channel 14/stable

# relate charms
juju relate postgresql-k8s:database trino-k8s:postgresql_db
````
## Expected behavior
A relation is created between `trino-k8s` charm and `postgresql-k8s` charm, resulting in `db_context` being logged for the admin user.

## Actual behavior
Postgres Charm throw an error and goes into error status

## Versions

Operating system: `Ubuntu 22.04.1 LTS`

Juju CLI: `3.1.2-genericlinux-amd64`

Juju agent: `3.1.2`

Charm revision: `14.7`

microk8s: `MicroK8s v1.25.10 revision 5354`

## Log output

Juju debug log: 
[log.txt](https://github.com/canonical/postgresql-k8s-operator/files/11834560/log.txt)

```
2023-06-22T11:32:29.318Z [container-agent] 2023-06-22 11:32:29 ERROR juju-log database:134: Failed to create user: syntax error at or near ";"
2023-06-22T11:32:29.318Z [container-agent] LINE 1: ...134" WITH LOGIN ENCRYPTED PASSWORD 'QN20iew3AUq51Yve' admin;
2023-06-22T11:32:29.318Z [container-agent]                                                                       ^
2023-06-22T11:32:29.318Z [container-agent] 
2023-06-22T11:32:29.324Z [container-agent] 2023-06-22 11:32:29 ERROR juju-log database:134: 
2023-06-22T11:32:29.324Z [container-agent] Traceback (most recent call last):
2023-06-22T11:32:29.324Z [container-agent]   File "/var/lib/juju/agents/unit-postgres-two-0/charm/lib/charms/postgresql_k8s/v0/postgresql.py", line 174, in create_user
2023-06-22T11:32:29.324Z [container-agent]     cursor.execute(
2023-06-22T11:32:29.324Z [container-agent] psycopg2.errors.SyntaxError: syntax error at or near ";"
[...]

PR #201, opened 2023-07-17
[DPE-2167] Added admin extra user role

marceloneppel:
## Issue
There is no `admin` extra user role that can be requested by client application charms.

## Solution
Add only read permissions to the special `admin` role. The permission to create anything in the public schema of the `postgres` database was removed.

Also add to it the permission to create and drop databases (except the system database `postgres`).

While reviewing, the following files can be ignored (they are generated):
* `poetry.lock`
* `requirements.txt`
* `tests/integration/ha_tests/application-charm/requirements.txt`
* `tests/integration/new_relations/application-charm/requirements.txt`

We disabled the hashes temporarily on all the `requirements.txt` files due to [canonical/charmcraft#1179] (so we can pack the charms now). I will be reverted later.

Fixes https://github.com/canonical/postgresql-k8s-operator/issues/192.


lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -178,28 +178,33 @@ def create_user(
...
+                    admin_role = "admin" in extra_user_roles
+                    roles = [role[0] for role in cursor.fetchall() if role[0] != "admin"]
+                    privileges = {
                         extra_user_role
                         for extra_user_role in extra_user_roles
-                        if extra_user_role not in roles
-                    ]
+                        if extra_user_role not in roles and extra_user_role != "admin"
+                    }
+                    if "SUPERUSER" in map(str.upper, privileges):
```
dragomirp, line 195: I think PGB is using superuser. We should change that to admin after this change.

dragomirp (approved):

lib/charms/postgresql_k8s/v0/postgresql.py
marceloneppel: Good point. I reverted this part of the code as I realised that Landscape needs a superuser (for plpython extension, a requirement that may or may not be removed at some time, but also due to some schema changes that may happen in a landscape server upgrade).

taurus-forever: 1) we should document this properly. JFMI, which extra_user_role used by pgbouncer?
2) re: `plpython`. We will have to hide it for now (c) Mohamed. Security concerns. This is why it was missing on mid-sync presentation. Mohamed will create a ticket to discuss and hide in documentation for now. No rush necessary today!

taurus-forever (approved):
Thank you!
What about to improve https://charmhub.io/postgresql-k8s/docs/e-users to expand "Relation/integration users" with description of extra_user_roles  (internal and allowed o be used "admin". BTW, if no `extra-user-roles` requested, how do we call this "role"? :-) )

lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -129,7 +129,7 @@ def create_database(self, database: str, user: str) -> None:
                     sql.Identifier(database)
                 )
             )
-            for user_to_grant_access in [user] + self.system_users:
+            for user_to_grant_access in [user, "admin"] + self.system_users:
```
delgod, line 132: Is this line needed if we add `pg_read_all_data, pg_write_all_data` to the role?

lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -178,28 +178,28 @@ def create_user(
...
                 # Create or update the user.
                 cursor.execute(f"SELECT TRUE FROM pg_roles WHERE rolname='{user}';")
                 if cursor.fetchone() is not None:
                     user_definition = "ALTER ROLE {}"
                 else:
                     user_definition = "CREATE ROLE {}"
-                user_definition += (
-                    f"WITH LOGIN{' SUPERUSER' if admin else ''} ENCRYPTED PASSWORD '{password}'"
-                )
+                user_definition += f"WITH LOGIN{' SUPERUSER' if admin else ''} ENCRYPTED PASSWORD '{password}'{'IN ROLE admin CREATEDB CREATEROLE' if admin_role else ''}"
```
delgod, line 202: 1. hm, is `CREATEROLE` needed for relation users? I think relation users with `extra_user_role=admin` shouldn't be able to create other users.
2. should we use `NOLOGIN` for `admin` role?
```suggestion
                user_definition += f"WITH {'NOLOGIN' if user == "admin" else 'LOGIN'}{' SUPERUSER' if admin else ''} ENCRYPTED PASSWORD '{password}'{'IN ROLE admin CREATEDB' if admin_role else ''}"
```

lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -347,15 +347,21 @@ def set_up_database(self) -> None:
         """Set up postgres database with the right permissions."""
         connection = None
         try:
+            self.create_user(
+                "admin",
+                extra_user_roles="pg_read_all_data,pg_read_all_settings,pg_read_all_stats,pg_stat_scan_tables,pg_monitor,pg_signal_backend",
```
delgod, line 352: I would limit privs by read and write all databases.
```suggestion
                extra_user_roles="pg_read_all_data,pg_write_all_data",
```

delgod (changes requested):

lib/charms/postgresql_k8s/v0/postgresql.py
marceloneppel: Good catch! I fixed both situations on https://github.com/canonical/postgresql-k8s-operator/pull/201/commits/3b9591933d20ef01dddf3429181ac2a2f9607e6b.

lib/charms/postgresql_k8s/v0/postgresql.py
marceloneppel: Updated on https://github.com/canonical/postgresql-k8s-operator/pull/201/commits/3b9591933d20ef01dddf3429181ac2a2f9607e6b.

lib/charms/postgresql_k8s/v0/postgresql.py
marceloneppel: It's still needed after we simplify the privileges to only those two. At least the `CONNECT` privilege is needed (which is granted through this block).

lib/charms/postgresql_k8s/v0/postgresql.py
marceloneppel: PgBouncer currently uses `SUPERUSER` (to be able to create other superusers).

marceloneppel:
> Thank you! What about to improve https://charmhub.io/postgresql-k8s/docs/e-users to expand "Relation/integration users" with description of extra_user_roles (internal and allowed o be used "admin". BTW, if no `extra-user-roles` requested, how do we call this "role"? :-) )

Good suggestion. I added the following text to both https://discourse.charmhub.io/t/charmed-postgresql-k8s-explanations-users/10843 and https://discourse.charmhub.io/t/charmed-postgresql-explanations-users/10798:

> When an application charm requests a new user through the relation/integration it can specify that the user should have the `admin` role in the `extra_user_roles` field. The `admin` role enables the new user to read and write to all databases (for the `postgres` system database it can only read data) and also to create and delete non-system databases.

delgod (approved):

PR #209, opened 2023-07-25
[DPE-1764] Juju 3 secrets

dragomirp:
## Issue
Charm doesn't use secrets for internal sensitive values

## Solution
Use Juju secrets instead of peer data if available
Updated the charm libs


renovate.json
```diff
@@ -33,7 +33,7 @@
       "matchDepNames": ["libjuju 3"],
       "matchPackageNames": ["juju"],
       "matchDatasources": ["pypi"],
-      "allowedVersions": "<3.2.0",
+      "allowedVersions": "<4.0.0",
```
dragomirp, line 36: libjuju should be pinned to the major version, not the minor.

src/charm.py
dragomirp, line 1: Mostly copy paste from the mongo charm.

src/charm.py
```diff
@@ -89,6 +100,8 @@ def __init__(self, *args):
         self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.leader_elected, self._on_leader_elected)
         self.framework.observe(self.on[PEER].relation_changed, self._on_peer_relation_changed)
+        self.framework.observe(self.on.secret_changed, self._on_peer_relation_changed)
+        self.framework.observe(self.on.secret_remove, self._on_peer_relation_changed)
```
dragomirp, line 104: To update configs.

src/charm.py
```diff
@@ -155,30 +168,167 @@ def unit_peer_data(self) -> Dict:
...
+
+    def _juju_secret_set(self, scope: str, key: str, value: str) -> str:
+        """Helper function setting Juju secret."""
+        if scope == UNIT_SCOPE:
+            peer_data = self.unit_peer_data
+        elif scope == APP_SCOPE:
+            peer_data = self.app_peer_data
+        self._juju_secrets_get(scope)
+
+        key = SECRET_KEY_OVERRIDES.get(key, key)
```
dragomirp, line 244: Secret key needs to be at least 3 chars, so ca needs to be changed to something longer.

tests/integration/test_tls.py
```diff
@@ -119,19 +136,9 @@ async def test_mattermost_db(ops_test: OpsTest) -> None:
...
             with attempt:
-                logger.info(f"checking if pg_rewind used TLS on {replica}")
-                logs = await run_command_on_unit(
-                    ops_test,
-                    replica,
-                    "grep rewind /var/log/postgresql/postgresql.log",
-                )
-                assert (
-                    "connection authorized: user=rewind database=postgres SSL enabled" in logs
-                ), "TLS is not being used on pg_rewind connections"
```
dragomirp, line 134: This check consistently failed with secrets enabled. Reworked to check all units.

dragomirp:
While this PR will use the Juju secret store, most of the passwords would still be written on disk in the patroni config file. We should most probably consider replacing what we can with [envvars](https://patroni.readthedocs.io/en/latest/ENVIRONMENT.html).

tests/integration/test_tls.py
marceloneppel: I think we can check only the current leader and sync-stanby, in where the connections should be done.

delgod (changes requested):

src/charm.py
```diff
@@ -1157,6 +1309,19 @@ def update_config(self) -> bool:
...
+        current_layer = container.get_plan()
+        if metrics_service := current_layer.services[self._metrics_service]:
+            if not metrics_service.environment.get("DATA_SOURCE_NAME", "").startswith(
+                f"user={MONITORING_USER} password={self.get_secret('app', MONITORING_PASSWORD_KEY)} "
+            ):
+                container.add_layer(
+                    self._metrics_service,
+                    Layer({"services": {self._metrics_service: self._generate_metrics_service()}}),
+                )
+                container.restart(self._metrics_service)
```
dragomirp, line 1323: If env is not correct, regenerate and restart.

src/charm.py
```diff
@@ -155,30 +168,165 @@ def unit_peer_data(self) -> Dict:
...
+
     def get_secret(self, scope: str, key: str) -> Optional[str]:
         """Get secret from the secret storage."""
-        if scope == "unit":
+        if scope not in [APP_SCOPE, UNIT_SCOPE]:
+            raise RuntimeError("Unknown secret scope.")
+
+        juju_version = JujuVersion.from_environ()
+
+        if juju_version.has_secrets:
```
delgod, line 227: if the user on juju 3.1.x updates the current `14/stable` charm (which doesn't support secrets) to this version (which supports secrets). I think the charm will become broken because not be able to fetch secrets from databag.

delgod (changes requested):

src/charm.py
dragomirp: Changed to first check databag.

delgod (approved):

src/charm.py
```diff
@@ -155,30 +168,181 @@ def unit_peer_data(self) -> Dict:
...
+
+        if scope == UNIT_SCOPE:
+            result = self.unit_peer_data.get(key, None)
+        else:
+            result = self.app_peer_data.get(key, None)
+
+        # TODO change upgrade to switch to secrets once minor version upgrades is done
+        if result:
+            return result
+
```
delgod, line 233: Problem with this implementation: we always take value from databag even if data exists in secret.
So, the following scenario is possible: the user updated the charm and rotated the password for the monitor user. Password rotated and written into a secret, but the password from databag was used anyway.

delgod (changes requested):

src/charm.py
dragomirp: The peer databag secret is deleted when setting or removing in Juju secrets, so we shouldn't be getting a stale values.

delgod (approved):

dragomirp:
Added [DPE-2324](https://warthogs.atlassian.net/browse/DPE-2324) to move the secrets during upgrade charm once we have both implementation.

marceloneppel (approved):

PR #219, opened 2023-08-10
[MISC] Switch to Ruff

dragomirp:
## Issue
Charmcraft templates default to ruff as a linter and keeping flake8 depends on some packages that appear unmaintained

## Solution
Switch to ruff while maintaining compatibility with preexisting checks


pyproject.toml
```diff
@@ -100,23 +94,39 @@ markers = ["unstable"]
...
+    "D215",
+    "D400",
+    "D404",
+    "D406",
+    "D407",
+    "D408",
+    "D409",
+    "D413",
+]
+# Ignore E501 because using black creates errors with this
```
dragomirp, line 116: Mostly copypaste from [ the charmcraft template](https://github.com/canonical/charmcraft/blob/main/charmcraft/templates/init-simple/pyproject.toml.j2#L17)

pyproject.toml
```diff
@@ -100,23 +94,39 @@ markers = ["unstable"]
...
-[tool.flake8]
-max-line-length = 99
-max-doc-length = 99
-max-complexity = 10
-exclude = [".git", "__pycache__", ".tox", "build", "dist", "*.egg_info", "venv"]
-select = ["E", "W", "F", "C", "N", "R", "D", "H"]
-# Ignore W503, E501 because using black creates errors with this
+[tool.ruff]
+target-version = "py38"
+src = ["src", "."]
```
dragomirp, line 100: We should most probably add `tests` as `known-local-folder`, but that will diverge from preexisting behaviour.

pyproject.toml
```diff
@@ -100,23 +94,39 @@ markers = ["unstable"]
...
 # Linting tools configuration
-[tool.flake8]
-max-line-length = 99
-max-doc-length = 99
-max-complexity = 10
-exclude = [".git", "__pycache__", ".tox", "build", "dist", "*.egg_info", "venv"]
-select = ["E", "W", "F", "C", "N", "R", "D", "H"]
-# Ignore W503, E501 because using black creates errors with this
+[tool.ruff]
+target-version = "py38"
```
dragomirp, line 99: Black is configured so and charm libs must be 3.8 compatible, but IMHO we can set 3.10 for pure jammy charms.

.github/workflows/release.yaml
```diff
@@ -13,6 +13,7 @@ on:
       - poetry.lock
       - pyproject.toml
       - '.github/**.md'
+      - '.github/workflows/ci.yaml'
```
dragomirp, line 16: Agents are pinned there.

pyproject.toml
```diff
@@ -31,26 +31,20 @@ optional = true
 
 [tool.poetry.group.format.dependencies]
 black = "23.7.0"
-isort = "5.12.0"
+ruff = "0.0.284"
```
dragomirp, line 34: Ruff can also format.

pyproject.toml
```diff
@@ -100,23 +94,39 @@ markers = ["unstable"]
...
-max-doc-length = 99
-max-complexity = 10
-exclude = [".git", "__pycache__", ".tox", "build", "dist", "*.egg_info", "venv"]
-select = ["E", "W", "F", "C", "N", "R", "D", "H"]
-# Ignore W503, E501 because using black creates errors with this
+[tool.ruff]
+target-version = "py38"
+src = ["src", "."]
+line-length = 99
+select = ["E", "W", "F", "C", "N", "D", "I001", "CPY001"]
```
dragomirp, line 102: We should most probably ad A (flake8-builtins), but that doesn't seem to have been working before either.

pyproject.toml
marceloneppel: Must charm libs be 3.8 compatible due to Juju 2.9 or something like that?

.github/workflows/release.yaml
dragomirp: This should only stop the release workflow. E.g. we bump the agent, we run the tests on the PR but don't release when we merge to main, since the charm didn't really change. In principle, it is possible to sneak breaking changes in the ignored files, but only for the tests. If the charm is changed, the release workflow will still run.

pyproject.toml
marceloneppel: So I think it's ok to keep the PG charm lib compatible with PgBouncer VM (which supports focal).

marceloneppel (approved):
LGTM! Thanks!

.github/workflows/release.yaml
taurus-forever: I am not 100% sure here, but let's try. It sounds safe.

pyproject.toml
taurus-forever: We have to keep focal compatibility at least!

pyproject.toml
taurus-forever: Yes, we should, consider to check the rabbit hole size in the follow-up PR.

taurus-forever (approved):
Excellent jump!

PR #220, opened 2023-08-10
[MISC] Missing await

dragomirp:
## Issue
Missing await in the tls tests causes frequent errors.


marceloneppel (approved):
Thanks!

taurus-forever (approved):
Nice catch!

PR #233, opened 2023-09-01
[DPE-2504] fully routable patroni cos

dragomirp:
## Issue
Cross model relation between COS Prometheus and Postgresql can't resolve the provided Patroni endpoint and it will not register new nodes.

## Solution
Switch to wildcard configuration.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -149,9 +149,7 @@ class PostgresqlOperatorCharm(CharmBase):
         return [
             {"static_configs": [{"targets": [f"*:{METRICS_PORT}"]}]},
             {
-                "static_configs": [
-                    {"targets": [f"{self.get_hostname_by_unit(self.unit.name)}:8008"]}
-                ],
+                "static_configs": [{"targets": [f"{ self._build_unit_name()}:8008"]}],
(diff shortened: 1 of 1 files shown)


dragomirp:
Wildcard is using pod IPs which should be routable between modesl, however that likely means we won't be able to verify the certificate from Prometheus.

taurus-forever (approved):
Thank you for the quick fix!

taurus-forever:
Q: why it is a problem with Patroni, and no such problem with PostgreSQL/PgBouncer? Are we exporting on Primary/Leader unit only for PgSQL/PgBouncer? Tnx!

dragomirp:
> Q: why it is a problem with Patroni, and no such problem with PostgreSQL/PgBouncer? Are we exporting on Primary/Leader unit only for PgSQL/PgBouncer? Tnx!

Patroni API is its own exporter, but it's also potentially served over TLS. Everything else uses a dedicated exporter which serve plaintext. So Patroni was the only thing that used hostnames (for potential cert validation) instead of wildcard IP handled by the COS lib internally.

PR #240, opened 2023-09-08
[DPE-2546] Split stanza create and stanza check

dragomirp:
## Issue
Pgbackrest stanza check requires  TLS servers to run if enabled, but they won't start until the hook is released and peer data is updated.

## Solution
Run only stanza create in the s3 relation hooks and trigger the stanza check via peer relation change


src/backups.py
```diff
@@ -299,12 +299,18 @@ def _initialise_stanza(self) -> None:
             self.charm.unit.status = BlockedStatus(FAILED_TO_INITIALIZE_STANZA_ERROR_MESSAGE)
             return
 
+        self.start_stop_pgbackrest_service()
+
         # Store the stanza name to be used in configurations updates.
-        self.charm.app_peer_data.update({"stanza": self.stanza_name})
+        self.charm.app_peer_data.update({"stanza": self.stanza_name, "init-pgbackrest": "True"})
```
dragomirp, line 305: An extra flag just in case.

src/backups.py
```diff
@@ -317,13 +323,14 @@ def _initialise_stanza(self) -> None:
...
             self.charm.app_peer_data.update({"stanza": ""})
+            self.charm.app_peer_data.pop("init-pgbackrest", None)
             self.charm.update_config()
 
             logger.exception(e)
             self.charm.unit.status = BlockedStatus(FAILED_TO_INITIALIZE_STANZA_ERROR_MESSAGE)
             return
 
-        return
+        self.charm.app_peer_data.pop("init-pgbackrest", None)
```
dragomirp, line 333: Clear the flag so we stop checking the stanza.

src/charm.py
```diff
@@ -474,7 +477,7 @@ def _on_peer_relation_changed(self, event: HookEvent) -> None:
             event.defer()
             return
         else:
-            self.unit_peer_data.pop("start-tls.server", None)
+            self.unit_peer_data.pop("start-tls-server", None)
```
taurus-forever, line 480: This is an interesting catch... 

Q: why do we set/remove it if we do not check it?
```
> ripgrep start-tls.server
src/charm.py
470:            self.unit_peer_data.update({"start-tls-server": "True"})
477:            self.unit_peer_data.pop("start-tls.server", None)
```

taurus-forever (approved):
LGTM, the idea to seaprat stanza init and check is a proper approach, IMHO.
Thank you!

src/charm.py
dragomirp: I'm unsure, so I left it be for the time being.

cjohnston1158:
This change resolved the issue for me.

src/charm.py
```diff
@@ -464,6 +464,9 @@ def _on_peer_relation_changed(self, event: HookEvent) -> None:
 
         self.postgresql_client_relation.update_read_only_endpoint()
 
+        if self.unit.is_leader() and "init-pgbackrest" in self.app_peer_data:
```
delgod, line 467: Should we move the condition inside the `check_stanza` function to not spread the `init-pgbackrest` logic between many different functions?

src/backups.py
```diff
@@ -299,12 +299,18 @@ def _initialise_stanza(self) -> None:
...
         # Store the stanza name to be used in configurations updates.
-        self.charm.app_peer_data.update({"stanza": self.stanza_name})
+        self.charm.app_peer_data.update({"stanza": self.stanza_name, "init-pgbackrest": "True"})
 
+    def check_stanza(self) -> None:
+        """Runs the pgbackrest stanza validation."""
         # Update the configuration to use pgBackRest as the archiving mechanism.
         self.charm.update_config()
 
+        self.charm.unit.status = MaintenanceStatus("checking stanza")
```
delgod, line 312: Do we really need this status? I think it is ok just put "Checking stanza" into logs.

delgod (approved):

src/backups.py
dragomirp: We don't need it, but I think it makes things clearer to the user.

marceloneppel (approved):

PR #243, opened 2023-09-08
[MISC] Clean up the artefacts in Patroni dashboard

dragomirp:
## Issue
The Patroni and Postgresql versions in the Patroni dashboard look weird

## Solution
Try to clean up


taurus-forever (approved):
Nice!

PR #250, opened 2023-09-12
[MISC] Temporarily disable indico tests on Juju 3

dragomirp:
## Issue
The indico legacy rel test is failing frequently on juju 3. Indico seems stuck in idle/active until the test times out or is cancelled.

## Solution
Temporarily disable the test.
Checking if things are better in juju 3.1.6 on https://github.com/canonical/postgresql-k8s-operator/pull/249

### DIFF
--- a/tests/integration/test_db.py
+++ b/tests/integration/test_db.py
@@ -4,6 +4,7 @@
 import logging
 from asyncio import gather
 
+import pytest
 from pytest_operator.plugin import OpsTest
 
 from tests.integration.helpers import (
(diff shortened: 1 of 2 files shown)


dragomirp:
Stability looks better in Juju 3.1.6, so I think we should be able to reenable that test when it hits stable.

tests/integration/test_db.py
```diff
@@ -229,6 +231,8 @@ async def test_indico_db_blocked(ops_test: OpsTest) -> None:
 
 async def test_discourse(ops_test: OpsTest):
     database_application_name = f"extensions-{DATABASE_APP_NAME}"
+    if database_application_name not in ops_test.model.applications:
+        await build_and_deploy(ops_test, 1, database_application_name)
```
dragomirp, line 235: Making sure it's deployed in case the preceding test is skipped.

marceloneppel (approved):
LGTM

taurus-forever (approved):

PR #251, opened 2023-09-14
[DPE-1755] Log rotation

dragomirp:
## Issue
No log rotation rules for Postgresql logs.

## Solution
Add log rotation rules from the spec.

### DIFF
--- a/src/constants.py
+++ b/src/constants.py
@@ -21,9 +21,9 @@ WORKLOAD_OS_GROUP = "postgres"
 WORKLOAD_OS_USER = "postgres"
 METRICS_PORT = "9187"
 POSTGRES_LOG_FILES = [
-    "/var/log/pgbackrest",
+    "/var/log/pgbackrest/*",
     "/var/log/postgresql/patroni.log",
-    "/var/log/postgresql/postgresql.log",
+    "/var/log/postgresql/postgresql*.log",
 ]
 # List of system usernames needed for correct work of the charm/workload.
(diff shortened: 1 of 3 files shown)


templates/patroni.yml.j2
dragomirp, line 1: Mostly copy paste from the spec.

src/constants.py
```diff
@@ -21,9 +21,9 @@
 WORKLOAD_OS_USER = "postgres"
 METRICS_PORT = "9187"
 POSTGRES_LOG_FILES = [
-    "/var/log/pgbackrest",
+    "/var/log/pgbackrest/*",
     "/var/log/postgresql/patroni.log",
-    "/var/log/postgresql/postgresql.log",
+    "/var/log/postgresql/postgresql*.log",
```
dragomirp, line 26: Globs for COS to be able to figure out the logs locations.

tests/integration/test_tls.py
```diff
@@ -47,7 +47,7 @@ async def check_tls_rewind(ops_test: OpsTest) -> None:
             logs = await run_command_on_unit(
                 ops_test,
                 unit.name,
-                "grep rewind /var/log/postgresql/postgresql.log",
+                "grep rewind /var/log/postgresql/postgresql-*.log",
```
taurus-forever, line 50: For the future: check it better as the current check does not guaranties a tls in use... it is just a rewind word on logs...

taurus-forever (approved):

PR #254, opened 2023-09-19
DPE-2616 Basic profile support

paulomach:
## Issue

Profiles support missing from PG

## Solution

* Created method to build postgresql parameters, based on profile.
* using patroni template to configure parameters
* defaults to `production` profiles (better performance, as done for mysql)
* parameter building is a MVP:
  * `testing` profile set no extra parameters
  * `production` set's minimal needs for IS scenarios (see follow-up)
  * has `available_memory` input, but is not in use: logic is TBD
* integration tests uses `testing` profile

## Folow-up PRs

1. Proper memory detection for parameter calculation
1. React `on-config-change` to allow inflight profile change

### DIFF
--- a/lib/charms/postgresql_k8s/v0/postgresql.py
+++ b/lib/charms/postgresql_k8s/v0/postgresql.py
@@ -19,7 +19,7 @@ The `postgresql` module provides methods for interacting with the PostgreSQL ins
 Any charm using this library should import the `psycopg2` or `psycopg2-binary` dependency.
 """
 import logging
-from typing import List, Set, Tuple
+from typing import List, Optional, Set, Tuple
 
 import psycopg2
 from psycopg2 import sql
@@ -32,7 +32,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 12
+LIBPATCH = 13
 
 INVALID_EXTRA_USER_ROLE_BLOCKING_MESSAGE = "invalid role(s) for extra user roles"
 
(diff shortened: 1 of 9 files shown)


src/charm.py
```diff
@@ -1313,6 +1313,7 @@ def update_config(self, is_creating_backup: bool = False) -> bool:
             backup_id=self.app_peer_data.get("restoring-backup"),
             stanza=self.app_peer_data.get("stanza"),
             restore_stanza=self.app_peer_data.get("restore-stanza"),
+            parameters=self.postgresql.build_postgresql_parameters(self.config["profile"], 0),
```
marceloneppel, line 1316: Should the user be able to change the profile after the charm was already deployed?

paulomach: Yes - it will need to happen - to keep it simple I will do on follow up PR.

marceloneppel (approved):
LGTM! Thanks Paulo!

lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -419,3 +419,20 @@ def update_user_password(self, username: str, password: str) -> None:
...
+        Args:
+            profile: the profile to use.
+            available_memory: available memory to use in calculation
+
+        Returns:
+            Dictionary with the PostgreSQL parameters.
+        """
+        logger.debug(f"Building PostgreSQL parameters for {profile=} and {available_memory=}")
+        if profile == "production":
+            return {"shared_buffers": "987MB", "effective_cache_size": "2958MB"}
```
paulomach, line 438: This static setup reflects the need from IS and will be replaced on next PR

dragomirp (approved):

PR #261, opened 2023-09-21
[DPE-2626][DPE-2627] Create bucket once and clear up blocked statuses

dragomirp:
## Issue
* All units try to create a bucket and all units will block if it fails
* Backup blocked statuses are not cleared if the s3 relation is removed.

## Solution
* Only the leader should check the s3 creds and bucket
* Clear up the blocked statuses when the s3 creds are gone

### DIFF
--- a/src/backups.py
+++ b/src/backups.py
@@ -34,6 +34,12 @@ FAILED_TO_ACCESS_CREATE_BUCKET_ERROR_MESSAGE = (
 )
 FAILED_TO_INITIALIZE_STANZA_ERROR_MESSAGE = "failed to initialize stanza, check your S3 settings"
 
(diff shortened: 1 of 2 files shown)


src/backups.py
```diff
@@ -369,6 +376,10 @@ def _on_s3_credential_changed(self, event: CredentialsChangedEvent):
             logger.debug("Cannot set pgBackRest configurations, missing configurations.")
             return
 
+        # Verify the s3 relation only on the leader
+        if not self.charm.unit.is_leader():
+            return
+
```
dragomirp, line 382: Do the init checks only on the leader unit. Followers should be updated from the peer change events.

taurus-forever (approved):
Nice!

PR #241, opened 2023-09-08
[WIP][POC][IGNORE][DPE-2568] relation secrets

juditnovak:
This is not a real PR, but space that allows to test the new version of `data-platform-libs` including Relation Secrets to be checked against Postgres tests.

PR #263, opened 2023-09-24
[DPE-2298] Relational secrets

dragomirp:
Continuation of #241 

Tweak tests to work with data_interfaces secrets.


tests/integration/new_relations/test_new_relations.py
```diff
@@ -87,6 +87,26 @@ async def test_database_relation_with_charm_libraries(ops_test: OpsTest, databas
         )
         await ops_test.model.wait_for_idle(apps=APP_NAMES, status="active", raise_on_blocked=True)
 
+        # Check that on juju 3 we have secrets and no username and password in the rel databag
+        if hasattr(ops_test.model, "list_secrets"):
```
dragomirp, line 91: If we have `list_secrets()` we're on juju 3

tests/integration/new_relations/test_new_relations.py
```diff
@@ -87,6 +87,26 @@ async def test_database_relation_with_charm_libraries(ops_test: OpsTest, databas
...
+                ),
+                get_application_relation_data(
+                    ops_test,
+                    APPLICATION_APP_NAME,
+                    FIRST_DATABASE_RELATION_NAME,
+                    "password",
+                ),
+            )
+            assert secret_uri is not None
+            assert password is None
```
dragomirp, line 108: We have secret uri and no password

tests/unit/conftest.py
dragomirp, line 1: Moved to unit dir so it doesn't mess with the integration tests.

tests/unit/conftest.py
```diff
@@ -0,0 +1,49 @@
...
+def juju_has_secrets(mocker: MockerFixture):
+    """This fixture will force the usage of secrets whenever run on Juju 3.x.
+
+    NOTE: This is needed, as normally JujuVersion is set to 0.0.0 in tests
+    (i.e. not the real juju version)
+    """
+    if juju_version := os.environ.get("LIBJUJU"):
+        juju_version = juju_version[1:].split(".")[0]
+    else:
+        juju_version = "3"
```
dragomirp, line 21: This should be changed to fixture params, but that doesn't work with `unittest.TestCase` suites. I propose refactoring that after the sprint.

tests/unit/test_charm.py
dragomirp, line 1: Changed to not assume databag secret backend.

tests/unit/test_backups.py
dragomirp, line 1: Not part of secrets, but ported over from VM.

tests/integration/new_relations/helpers.py
```diff
@@ -39,12 +48,25 @@ async def build_connection_string(
...
     if database is None:
         database = f'{application_name.replace("-", "_")}_{relation_name.replace("-", "_")}'
-    username = await get_application_relation_data(
-        ops_test, application_name, relation_name, "username", relation_id, relation_alias
-    )
-    password = await get_application_relation_data(
-        ops_test, application_name, relation_name, "password", relation_id, relation_alias
-    )
+
+    if secret_uri := await get_application_relation_data(
```
dragomirp, line 52: Always check for secret uri, so that we don't have to keep track of charms that have not switched to the secrets implementation (data-integrator).

delgod (approved):

marceloneppel (approved):

tests/unit/conftest.py
taurus-forever: Please create a backlog ticket for the refactoring (to do not forget in a rush).

taurus-forever (approved):
Nice!

tests/integration/new_relations/test_new_relations.py
juditnovak: Why not `JujuVersion.has_secrets`... Just for consistency with the generic concept of checking for Juju3?

Issue #262, opened 2023-09-21
After restoring a backup, replication doesn't work


cjohnston1158:
As discussed on MM, after restoring a backup and scaling postgresql-k8s back up, replication doesn't happen.

marceloneppel:
Additional details below.

Steps to reproduce:

```sh
juju deploy postgresql-k8s --channel 14/edge --trust
juju deploy s3-integrator
# wait for the units to become active
juju run-action s3-integrator/0 sync-s3-credentials access-key=XXX secret-key=XXX --wait
juju config s3-integrator bucket=XXX path=XXX region=XXX
juju relate postgresql-k8s s3-integrator
# wait for stanza to be initialised
juju scale-application postgresql-k8s 3
# wait for the new units to become active
juju scale-application postgresql-k8s 1
# wait for the units to be removed
juju run-action postgresql-k8s/0 create-backup --wait
juju run-action postgresql-k8s/0 list-backups --wait
juju run-action postgresql-k8s/0 restore backup-id=XXX --wait
# wait for the restore to complete
juju scale-application postgresql-k8s 3
# wait for the new units to become active
curl {unit-ip}:8008/cluster
# we should see a big lag in the replicas
```

It seems that it's failing due to the storage being reused (so we have two different timelines).

taurus-forever:
re: `It seems that it's failing due to the storage being reused`

If the assumption is true (pretty logical), the workaround is simple:

```
juju run-action postgresql-k8s/0 restore backup-id=XXX --wait
# wait for the restore to complete
juju remove-storage pgdata/2 # scaled-down above postgresql-k8s/2
juju remove-storage pgdata/1 # scaled-down above postgresql-k8s/1
juju scale-application postgresql-k8s 3
```
It can be documented as a proper approach for now.

@marceloneppel can (should?) we detect this situation in charm and drop local timeline forcing re-sync from primary postgresql-k8s/0? CC: @delgod

PR #273, opened 2023-09-29
[DPE-2657] Fix replication after restore

marceloneppel:
## Issue
If we have a cluster of two or more units, scale it down to 1 unit, create a backup, restore the backup and scale up back again, the replication stop working due to reuse of the storage.

## Solution
In that situation, as we see a huge replication lag, we need to reinitialise the replicas.

Fixes https://github.com/canonical/postgresql-k8s-operator/issues/262.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -464,14 +464,13 @@ class PostgresqlOperatorCharm(CharmBase):
 
         # Restart the workload if it's stuck on the starting state after a timeline divergence
         # due to a backup that was restored.
-        if (
-            not self.is_primary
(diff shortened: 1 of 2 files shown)


taurus-forever (approved):
LGTM, will it work on demo installations if there are no traffic?

marceloneppel:
> LGTM, will it work on demo installations if there are no traffic?

Yes. The lag appears even on a demo installation due to the divergence in the timelines.

dragomirp (approved):

PR #282, opened 2023-10-05
[DPE-2640] patroni logs

dragomirp:
## Issue
Patroni log rotation

## Solution
It doesn't seem to be possible to rotate the logs every minute using patroni alone


dragomirp:
Example logs:
```
2023-10-05 13:44:49 UTC [15]: INFO: no action. I am (postgresql-k8s-0), the leader with the lock 
2023-10-05 13:44:59 UTC [15]: INFO: no action. I am (postgresql-k8s-0), the leader with the lock 
2023-10-05 13:45:09 UTC [15]: INFO: no action. I am (postgresql-k8s-0), the leader with the lock 
```

templates/patroni.yml.j2
```diff
@@ -33,7 +33,7 @@ bootstrap:
         log_statement_sample_rate: 1
         log_statement_stats: 'off'
         log_temp_files: 1
-        log_timezone: 'GMT'
+        log_timezone: 'UTC'
```
dragomirp, line 36: Changed after the discussion in the office hours.

templates/patroni.yml.j2
```diff
@@ -61,6 +61,10 @@ bootstrap:
 bypass_api_service: true
 log:
   dir: /var/log/postgresql
+  dateformat: "%Y-%m-%d %H:%M:%S %Z"
+  format: "%(asctime)s [%(process)d]: %(levelname)s: %(message)s "
```
dragomirp, line 65: Adding the PID

templates/patroni.yml.j2
```diff
@@ -61,6 +61,10 @@ bootstrap:
 bypass_api_service: true
 log:
   dir: /var/log/postgresql
```
dragomirp, line 63: Should the Patroni logs be moved to a separate dir?

templates/patroni.yml.j2
```diff
@@ -61,6 +61,10 @@ bootstrap:
...
+  dateformat: "%Y-%m-%d %H:%M:%S %Z"
+  format: "%(asctime)s [%(process)d]: %(levelname)s: %(message)s "
+  file_num: 10
+  file_size: 10485760
```
marceloneppel, line 67: We can probably reduce this to 600 bytes, so it should rotate the logs close to every minute (with the default app name - `postgresql-k8s` - most part of the messages have 91 bytes on primary and 121 bytes on replica, and they are sent to the log file every 10 seconds - according to the value of the `loop_wait` parameter). It's not an ideal solution, but it's better to make it rotate the logs more often.

taurus-forever: It smells a bit, but up to @7annaba3l as a feature requester.

templates/patroni.yml.j2
taurus-forever: It can stay here, IMHO.
Reasoning: patroni launches postgresql => newbies will benefits noticing patroni.log IF postgresql failed to start.

taurus-forever (approved):

templates/patroni.yml.j2
7annaba3l: Hi @dragomirp @taurus-forever @marceloneppel 
Let's go for the proposal from @marceloneppel . 
We will revisit this in the future and probably cleanup and align the config to the server part using logrotate ...

marceloneppel (approved):

PR #283, opened 2023-10-05
DPE-2656 Profile limite memory config support

paulomach:
## Issue

* **main issue**: no fine tune for memory allocation
* configurations are being set without validation
* configurations changes are not triggering changes in the workload

## Solution

* Add `profile-limit-memory` config option
* Usage of data platform `data-models` lib for config validation
* Usage of `rollingOps` lib for workload restart orchestration, when the changed configs requires.

### Todo

- proper limit memory usage
- figure out why leader unit keeps pending restart state after restart


lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -420,19 +421,39 @@ def update_user_password(self, username: str, password: str) -> None:
...
 
         Args:
             profile: the profile to use.
             available_memory: available memory to use in calculation in bytes.
+            limit_memory: (optional) limit memory to use in calculation in bytes.
 
         Returns:
             Dictionary with the PostgreSQL parameters.
         """
+        if limit_memory:
```
paulomach, line 455: This should be fixed in the future since it's not limiting the amount of memory used by postgres, but using the limit value in place of the available one (case smaller).

requirements.txt
dragomirp, line 1: Postgresql doesn't pin charmcraft so you should export without hashes.

src/config.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python3
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Structured configuration for the Kafka charm."""
```
dragomirp, line 5:
```suggestion
""Structured configuration for the Postgresql charm."""
```

paulomach: ops.. e845865

requirements.txt
paulomach: I've used the tox target to update it, so modified it her: e845865

dragomirp (approved):

marceloneppel (approved):
LGTM!

taurus-forever (approved):
Nice!

PR #290, opened 2023-10-15
[DPE-2569] Extensions not enabled when database is created

dragomirp:
## Issue
New databases do not consider already enabled extensions.

## Solution
Check extensions conf when creating a database and enable extensions.

### DIFF
--- a/lib/charms/postgresql_k8s/v0/postgresql.py
+++ b/lib/charms/postgresql_k8s/v0/postgresql.py
@@ -117,12 +117,13 @@ class PostgreSQL:
         connection.autocommit = True
         return connection
 
-    def create_database(self, database: str, user: str) -> None:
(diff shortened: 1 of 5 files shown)


lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -170,6 +171,10 @@ def create_database(self, database: str, user: str) -> None:
             logger.error(f"Failed to create database: {e}")
             raise PostgreSQLCreateDatabaseError()
 
+        # Enable preset extensions
+        for plugin in plugins:
+            self.enable_disable_extension(plugin, True, database)
+
```
dragomirp, line 177: We should consider for a way to get the plugins in PGB charms.

marceloneppel (approved):

taurus-forever (approved):

PR #297, opened 2023-10-20
[DPE-2569] Double check that extensions are enabled/disabled

dragomirp:
## Issue
Indico charm still failing to bootstrap consistently using the new interface.

## Solution
Double check that extensions are enabled in the internal lib.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -51,20 +51,20 @@ jobs:
       - name: Run tests
         run: tox run -e unit-${{ matrix.juju-version }}
 
-  lib-check:
-    name: Check libraries
-    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


marceloneppel (approved):

taurus-forever (approved):
I wish to have synchronous extension enabling approach and not checking their availability after enabling. Hovewer I do not see such an option here...

I believe the legacy interface is also affected here but luckly juju is slower in `database-change` even than PostgreSQL in enabling extention => we do not see it for legacy interface...

The possible workaround for indico is "sleep(10)" before trying to connect PostgreSQL.

dragomirp:
Further investigation implies that the issue is not in Postgresql extensions. Closing the PR for the time being.

PR #303, opened 2023-10-25
[DPE-2838] Secret labels

juditnovak:
Secrets with labels (refacored to a better shape)


lib/charms/postgresql_k8s/v0/postgresql_secrets.py
```diff
@@ -0,0 +1,141 @@
...
+        self.charm = charm
+        self._secrets: Dict[str, CachedSecret] = {}
+
+    def get(self, label: str, uri: Optional[str] = None) -> Optional[CachedSecret]:
+        """Getting a secret from Juju Secret store or cache."""
+        if not self._secrets.get(label):
+            secret = CachedSecret(self.charm, label, uri)
+            if secret.meta:
+                self._secrets[label] = secret
+        return self._secrets.get(label)
```
dragomirp, line 128:
```suggestion
        if not (secret := self._secrets.get(label)):
            secret = CachedSecret(self.charm, label, uri)
            if secret.meta:
                self._secrets[label] = secret
        return secret
```
No need to get twice.

src/charm.py
```diff
@@ -191,185 +190,101 @@ def unit_peer_data(self) -> Dict:
...
-
-        # TODO change upgrade to switch to secrets once minor version upgrades is done
-        if key in peer_data:
-            del peer_data[key]
+                return
 
-        return self.secrets[scope][SECRET_LABEL].id
+            value = secret.get_content().get(secret_key)
+            if value != SECRET_DELETED_LABEL:
+                return value
```
dragomirp, line 233: Why not do this in `self.secrets.get()`?

src/charm.py
```diff
@@ -191,185 +190,101 @@ def unit_peer_data(self) -> Dict:
...
-            self.unit_peer_data.update({key: value})
+            secret_key = self._translate_field_to_secret_key(key)
+            label = generate_secret_label(self, scope)
+            secret = self.secrets.get(label)
+            if not secret:
+                self.secrets.add(label, {secret_key: value}, scope)
+            else:
+                content = secret.get_content()
+                content.update({secret_key: value})
+                secret.set_content(content)
```
dragomirp, line 257: Why not handle this in `self.secret.add()`?

src/charm.py
```diff
@@ -191,185 +190,101 @@ def unit_peer_data(self) -> Dict:
...
-        if scope not in [APP_SCOPE, UNIT_SCOPE]:
+        if scope not in get_args(Scopes):
             raise RuntimeError("Unknown secret scope.")
 
         if not value:
             return self.remove_secret(scope, key)
 
-        juju_version = JujuVersion.from_environ()
+        if JujuVersion.from_environ().has_secrets:
+            if self._peer_data(scope).get(key):
```
dragomirp, line 244:
```suggestion
            if key in self._peer_data(scope):
```

src/charm.py
```diff
@@ -191,185 +190,101 @@ def unit_peer_data(self) -> Dict:
...
+                # Charm must have been upgraded since last run
+                # We move from databag to secrets
+                self._peer_data(scope).pop(key)
```
dragomirp, line 247:
```suggestion
            # Charm must have been upgraded since last run
            # We move from databag to secrets
            self._peer_data(scope).pop(key, None)
```

src/charm.py
```diff
@@ -191,185 +190,101 @@ def unit_peer_data(self) -> Dict:
...
+                return
 
-        juju_version = JujuVersion.from_environ()
-        if juju_version.has_secrets:
-            return self._juju_secret_remove(scope, key)
-        if scope == UNIT_SCOPE:
-            del self.unit_peer_data[key]
+            content[secret_key] = SECRET_DELETED_LABEL
+            secret.set_content(content)
         else:
```
dragomirp, line 283: We should always pop here, secret might have been originally set in the databag and never moved to secrets.

tests/integration/helpers.py
```diff
@@ -412,18 +412,28 @@ async def get_leader_unit(ops_test: OpsTest, app: str) -> Optional[Unit]:
...
     for unit in ops_test.model.applications[database_app_name].units:
-        if unit.name != down_unit:
-            action = await unit.run_action("get-password", **{"username": username})
-            result = await action.wait()
-            return result.results["password"]
+        if unit.name == down_unit:
+            continue
+
+        if pw := get_password_on_unit(ops_test, username, database_app_name, unit.name):
+            return pw
```
dragomirp, line 436:
```suggestion
        return await get_password_on_unit(ops_test, username, database_app_name, unit.name)
```
If it's None we'll return None either way.

src/charm.py
juditnovak: It's a very logical suggestion.
The original idea is/was to keep this module consistent across charms (hoping for later unification, i.e. central source).
For this reason I would rather suggest not to change the behavior and interface for now.

If no centralization happens, U could indeed do this in a later refactor.
Would that be acceptable 4 U?

dragomirp: Fine by me.

src/charm.py
juditnovak: I was thinking about this one too.

As long as hoping for unification of the `<charmnam>_secrets` module across charms, I'd like to leave this out.

The reason is: this is a Juju 3.1.6 > workaround.
Now we already have a bunch of charms that could drop Juju2. I'm wondering that they may be able 2 drop 3.1.5 at some point -- while other charms may still need to support it.

I'd rather not put a version-bound workaround in a lib, that has a chance for centralization.

It's a logical suggestion, but I'm wondering that --for now-- it may be better to keep it outside.
(If the moduel relamins local and OK to diverge, this could be refactored into the module)

tests/integration/helpers.py
juditnovak: :facepalm:

lib/charms/postgresql_k8s/v0/postgresql_secrets.py
juditnovak: So.... We actually have to.

The thing is that: on line 126 it may turn out that the secret actually doesn't exist.
In which case we're not adding it to the cache.

So we don't necessarily wanna return `secret`, only if it is linked with a real, existing thing.

I'm adding a line of comment on this, coz it's horribly implicit. Took me some time to figure it out after having got 15 unittests failing on this.

tests/integration/test_password_rotation.py
```diff
@@ -77,3 +77,49 @@ async def test_password_rotation(ops_test: OpsTest):
         if not await unit.is_leader_from_status():
             await restart_patroni(ops_test, unit.name)
             assert await check_patroni(ops_test, unit.name, restart_time)
+
+
+async def test_password_from_secret_same_as_cli(ops_test: OpsTest):
```
dragomirp, line 82:
```suggestion
@pytest.mark.juju3
async def test_password_from_secret_same_as_cli(ops_test: OpsTest):
```
This test can't run on Juju 2.9

dragomirp:
Upgrade tests seem to be failing because existing secrets on `14/edge` and `14/stable` are not labelled and after the upgrade the charm can no longer find them.

dragomirp (approved):

src/charm.py
```diff
@@ -191,185 +191,116 @@ def unit_peer_data(self) -> Dict:
...
+                return
 
-        return self.secrets[scope][SECRET_LABEL].id
+            secret_key = self._translate_field_to_secret_key(key)
+            value = secret.get_content().get(secret_key)
+            if value != SECRET_DELETED_LABEL:
+                return value
 
-    def set_secret(self, scope: str, key: str, value: Optional[str]) -> Optional[str]:
+    def set_secret(self, scope: Scopes, key: str, value: Optional[str]) -> Optional[str]:
```
delgod, line 250: lets move most of logic to lib in the future PR

delgod (changes requested):

delgod (approved):

PR #322, opened 2023-11-12
[DPE-2903] Toggle plugins in one go

dragomirp:
## Issue
Starting a connection for each db to enable or disable plugins will take more and more time as the number of plugins increases.

## Solution
Open a single connection per db and enable or disable the plugins in one go.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -552,24 +552,22 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             database: optional database where to enable/disable the extension.
         """
         original_status = self.unit.status
+        extensions = {}
+        # collect extensions
         plugins_exception = {"uuid_ossp": '"uuid-ossp"'}
         for plugin in self.config.plugin_keys():
             enable = self.config[plugin]
 
             # Enable or disable the plugin/extension.
             extension = "_".join(plugin.split("_")[1:-1])
-            if extension in plugins_exception:
-                extension = plugins_exception[extension]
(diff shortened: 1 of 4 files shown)


src/charm.py
```diff
@@ -552,24 +552,22 @@ def enable_disable_extensions(self, database: str = None) -> None:
...
-            try:
-                self.postgresql.enable_disable_extension(extension, enable, database)
-            except PostgreSQLEnableDisableExtensionError as e:
-                logger.exception(
-                    f"failed to {'enable' if enable else 'disable'} {extension} plugin: %s", str(e)
-                )
-            self.unit.status = original_status
+            extension = plugins_exception.get(extension, extension)
+            extensions[extension] = enable
+        self.unit.status = WaitingStatus("Updating extensions")
```
dragomirp, line 565: We can no longer tell if we are enabling or disabling and which plugin.

src/relations/db.py
```diff
@@ -152,9 +152,6 @@ def set_up_relation(self, relation: Relation) -> bool:
 
             self.charm.postgresql.create_database(database, user, plugins=plugins)
 
-            # Enable/disable extensions in the new database.
-            self.charm.enable_disable_extensions(database)
-
```
dragomirp, line 157: This will be done anyway on line 153

taurus-forever (approved):
LGTM. Can we improve the test to make sure at least some extension(s) are enabled?

marceloneppel (approved):
LGTM! Thanks for the great improvement!

PR #324, opened 2023-11-16
[DPE-2844] New interface indico test and new legacy extensions test

dragomirp:
## Issue
* Indico charm was ported to the new interface
* No tests left checking for legacy interface extension behaviour

## Solution
* Split the old indico test into one checking legacy relations and one testing that indico deploys with the new interface

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,7 +22,7 @@ on:
 jobs:
   lint:
     name: Lint
-    uses: canonical/data-platform-workflows/.github/workflows/lint.yaml@v5.1.2
+    uses: canonical/data-platform-workflows/.github/workflows/lint.yaml@v6.1.1
 
   unit-test:
     name: Unit test charm
@@ -45,7 +45,7 @@ jobs:
 
   build:
     name: Build charm
-    uses: canonical/data-platform-workflows/.github/workflows/build_charms_with_cache.yaml@v5.1.2
+    uses: canonical/data-platform-workflows/.github/workflows/build_charms_with_cache.yaml@v6.1.1
     permissions:
       actions: write  # Needed to manage GitHub Actions cache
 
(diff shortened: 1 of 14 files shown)


tests/integration/ha_tests/helpers.py
```diff
@@ -30,15 +28,14 @@
...
+    APPLICATION_NAME,
     app_name,
     db_connect,
     get_password,
     get_primary,
     get_unit_address,
 )
 
-APPLICATION_NAME = "postgresql-test-app"
-METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
```
dragomirp, line 41: We have those in the root helpers module

tests/integration/new_relations/test_new_relations.py
dragomirp, line 1: There's no special behaviour for extensions in the new interface, so just deploying indico and waiting to idle.

.github/workflows/ci.yaml
```diff
@@ -115,14 +115,8 @@ jobs:
             echo Skipping unstable tests
             echo "mark_expression=and not unstable" >> "$GITHUB_OUTPUT"
           fi
-      - name: Select test secret usage
-        id: select-test-secrets
-        if: ${{ github.event.pull_request.head.repo.full_name != 'canonical/postgresql-k8s-operator' }}
-        run: |
-          echo Skipping tests using secrets
-          echo "mark_secrets=and not uses_secrets" >> "$GITHUB_OUTPUT"
```
dragomirp, line 123: New dpw would skip automatically so we no longer need this logic.

tests/integration/test_db.py
dragomirp, line 1: The test app has a legacy interface with extensions for PGB tests, so using that to replace indico.

tests/integration/new_relations/test_new_relations.py
```diff
@@ -569,3 +569,41 @@ async def test_invalid_extra_user_roles(ops_test: OpsTest):
             raise_on_blocked=False,
             timeout=1000,
         )
+
+
+async def test_indico_db_blocked(ops_test: OpsTest) -> None:
```
marceloneppel, line 574: We can now remove the `blocked` word from the test name.

tests/integration/test_db.py
```diff
@@ -85,148 +88,97 @@ async def test_finos_waltz_db(ops_test: OpsTest) -> None:
...
+    logger.info("Verifying that the charm doesn't block when the extensions are enabled")
+    config = {"plugin_pg_trgm_enable": "True", "plugin_unaccent_enable": "True"}
+    await ops_test.model.applications[DATABASE_APP_NAME].set_config(config)
+    await ops_test.model.wait_for_idle(apps=[DATABASE_APP_NAME], status="active")
+    await ops_test.model.relate(f"{DATABASE_APP_NAME}:db", f"{APPLICATION_NAME}:db")
+    await ops_test.model.wait_for_idle(
+        apps=[DATABASE_APP_NAME, APPLICATION_NAME],
+        status="active",
+        raise_on_blocked=False,
+        timeout=2000,
```
marceloneppel, line 157: Can we change `raise_on_blocked` to `True` here and in the other places where the charm shouldn't block? It can speedup the test if something is wrong.

marceloneppel (approved):
LGTM! Nice split of tests.

tests/integration/new_relations/test_new_relations.py
dragomirp: Updated.

tests/integration/test_db.py
dragomirp: I think we don't really need it in any of the `wait_for_idle` calls (we track it with `block_until`).

taurus-forever (approved):

Issue #335, opened 2023-11-27
[Backup/Restore] Unable to restore backup in another cluster


javierdelapuente:
## Steps to reproduce

1. Backup database to a S3 bucket.
2. Create a new different cluster. Same controller and model names as in the previous one.
3. juju deploy postgresql-k8s --trust
4. juju deploy s3-integrator
5. juju run s3-integrator/leader sync-s3-credentials access-key=<access-key> secret-key=<secret-key> 
6. juju config s3-integrator .... (same as in the original cluster)
7. juju relate s3-integrator postgresql-k8s
At this point, with s3 correctly configured, the workload gets into `blocked` state with message `failed to initialize stanza, check your S3 settings`

8. juju run postgresql-k8s/leader list-backups
Backups are correctly shown.
9. juju run postgresql-k8s/leader restore backup-id=2023-11-27T09:37:51Z
Fails to restore backups, as the charm does not expect the current block state.

## Expected behavior
The backup should be restored in the new cluster.

## Versions

Operating system: 
Ubuntu 22.04.3 LTS

Juju CLI: 
3.1.6-genericlinux-amd64

Juju agent: 
```
Model             Controller  Cloud/Region        Version  SLA          Timestamp
tutorial-synapse  microk8s    microk8s/localhost  3.1.6    unsupported  15:26:31+01:00

App             Version  Status   Scale  Charm           Channel    Rev  Address         Exposed  Message
postgresql-k8s  14.9     waiting      1  postgresql-k8s  14/stable  158  10.152.183.134  no       installing agent
s3-integrator            active       1  s3-integrator   stable      13  10.152.183.70   no       

Unit               Workload  Agent  Address     Ports  Message
postgresql-k8s/0*  blocked   idle   10.1.42.34         failed to initialize stanza, check your S3 settings
s3-integrator/0*   active    idle   10.1.42.35         
```

Charm revision: 

microk8s: 
MicroK8s v1.27.7 revision 6101

## Log output

Juju debug log: 

When creating the integration:
```
unit-postgresql-k8s-0: 15:22:15 INFO unit.postgresql-k8s/0.juju-log s3-parameters:9: Bucket backups-bucket exists.
[...]

PR #354, opened 2024-01-09
[DPE-3061] Check system identifier in stanza

marceloneppel:
## Issue
In https://github.com/canonical/postgresql-k8s-operator/issues/335, the new cluster where the user tried to restore the backup had the same name as the cluster the backup was made from. As we're checking only the cluster/stanza name to set the blocked status, informing the user that there are backups from another cluster in the repository, the blocked status is not set for this specific case.

## Solution
Use the PostgreSQL installation system id to differentiate clusters.


dragomirp (approved):

taurus-forever (approved):
Tnx!

PR #358, opened 2024-01-15
[DPE-3199] Update secrets implementation

dragomirp:
Update the charm to the secrets implementation in the VM charm.


src/charm.py
```diff
@@ -1119,6 +1094,8 @@ def is_primary(self) -> bool:
     @property
     def is_tls_enabled(self) -> bool:
         """Return whether TLS is enabled."""
+        if not self.model.get_relation(PEER):
+            return False
```
dragomirp, line 1098: Used in the charm's init to set the monitoring services, which fails with the current implementation when the charm initially bootstraps.

juditnovak: Q: Wasn't it enough to prevent this in the `(get|set)_secret()` methods?
I came across the same issue on VM, thus I've added the same check to `(get|set)_secret()` if I recall right. I'm wondering shouldn't we rather add it to those functions instead?

I mean theoretically critical code should only be executeable from those two functions -- and then that should be the "central" source for this safety measure.

juditnovak (commented):
Could we pls-pls copy paste all those unittests I've added for secrets for VM?

https://github.com/canonical/postgresql-operator/pull/314/files#diff-0580ab9aef31b44580478eb270ec3422c5e6fef35df585fecb3a68317e246809

I do rely on those whenever testing secrest modifications...

marceloneppel (approved):

src/charm.py
dragomirp: Discussed with @juditnovak. The check is in the Mongo charms. I would prefer to keep this check outside of the secrets methods, as it is a edge case that shouldn't apply to most calls of those methods.

juditnovak (approved):
I added the unittests, good 2 go :-)

PR #369, opened 2024-01-22
[DPE-2900] Cross-region async replication integration tests

marceloneppel:
## Issue
There are no integration tests for cross-region async-replication.

## Solution
Add integration tests for some use cases:
- Establish async replication between two PostgreSQL clusters.
- Switchover to the secondary cluster.
- Promote a standby cluster after the relation is broken.
- Disable and reestablish async replication between two PostgreSQL clusters.
- Crash the `sync_standby` member of the main cluster and check that replication is still working.
- Crash the `standby_leader` member of the secondary cluster and check that replication is still working.


marceloneppel:
It was superseded by https://github.com/canonical/postgresql-k8s-operator/pull/448.

PR #404, opened 2024-02-28
Cover missing test cases for internal users

TakoB222:
### Issue
All this cases should be covered:
- Authentification should be enabled by default always. No anonymous/guest 
access ever;
- get-password and set-password functions should be created for internal 
users;


marceloneppel (approved):
LGTM!

dragomirp (approved):

taurus-forever (approved):
Nice! Thank you!

P.S. I would still add one testcase (can be separate PR): make sure all passwords are different (for different users). I mean the current test will pass IF all passwords are updated simultaneously for all users (due to some bug in code).

taurus-forever:
@TakoB222 please copy the test for VM charm as well. Thank you!

PR #416, opened 2024-03-10
Add `test_config_parameters` test

TakoB222:
### Issue
All this cases should be covered:
- Charge each charm config option using clusterized app and make sure changes are 
applied without error (and without downtime if possible):
  - check values boundaries (e.g. try 42 where 1-10 are allowed)
  - check types (e.g. try 42 where boolean true/false are expected)
  - check invalid config values (see the “test string” below)

### DIFF
--- a/tests/integration/test_config.py
+++ b/tests/integration/test_config.py
@@ -8,9 +8,9 @@ import pytest as pytest
 from pytest_operator.plugin import OpsTest
 
 from .helpers import (
-    CHARM_SERIES,
     DATABASE_APP_NAME,
-    get_leader_unit, build_and_deploy,
+    build_and_deploy,
+    get_leader_unit,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


tests/integration/test_config.py
```diff
@@ -0,0 +1,100 @@
...
+    """Build and deploy one unit of PostgreSQL and then test config with wrong parameters."""
+    # Build and deploy the PostgreSQL charm.
+    async with ops_test.fast_forward():
+        charm = await ops_test.build_charm(".")
+        await ops_test.model.deploy(
+            charm,
+            num_units=1,
+            series=CHARM_SERIES,
+            config={"profile": "testing"},
+        )
```
dragomirp, line 31: The k8s charm has slightly different parameters (trust, resource). Better to use the [build_and_deploy](https://github.com/canonical/postgresql-k8s-operator/blob/main/tests/integration/helpers.py#L57) helper

taurus-forever (approved):
Thank you! LGTM. P.S. K8s looks identical to VM here.
@dragomirp we should start think to extract common tests to test lib :-)

marceloneppel (approved):

PR #424, opened 2024-03-18
[MISC] Switch to ruff formatter

dragomirp:
Switch to ruff formatter

### DIFF
--- a/lib/charms/postgresql_k8s/v0/postgresql.py
+++ b/lib/charms/postgresql_k8s/v0/postgresql.py
@@ -18,6 +18,7 @@ The `postgresql` module provides methods for interacting with the PostgreSQL ins
 
 Any charm using this library should import the `psycopg2` or `psycopg2-binary` dependency.
 """
+
 import logging
 from collections import OrderedDict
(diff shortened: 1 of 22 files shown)


tests/integration/test_backups.py
```diff
@@ -90,12 +90,6 @@ async def cloud_configs(ops_test: OpsTest, github_secrets) -> None:
             bucket_object.delete()
 
 
-@pytest.mark.group(1)
-async def test_none() -> None:
-    """Empty test so that the suite will not fail if all tests are skippedi."""
-    pass
-
-
```
dragomirp, line 98: The other tests are skipped so we no longer need this.

lib/charms/postgresql_k8s/v0/postgresql.py
dragomirp, line 1: Black linting would fail on the lib. It should be fine, since it's internal for PG and PGB charms.

taurus-forever (approved):
Tnx!

marceloneppel (approved):

PR #425, opened 2024-03-19
Check user existence after relation broken for `db` and `db-admin` interfaces

TakoB222:
### Issue
User should be created on a relation-joined event, and the user should be dropped on a relation-departed event

### Solution
Added missed for checking user existence after application removed

### DIFF
--- a/tests/integration/test_db.py
+++ b/tests/integration/test_db.py
@@ -81,13 +81,14 @@ async def test_finos_waltz_db(ops_test: OpsTest) -> None:
             ANOTHER_FINOS_WALTZ_APP_NAME, block_until_done=True
         )
 
-        another_finos_waltz_users = []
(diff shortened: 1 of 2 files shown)


tests/integration/test_db.py
```diff
@@ -81,14 +81,15 @@ async def test_finos_waltz_db(ops_test: OpsTest) -> None:
             ANOTHER_FINOS_WALTZ_APP_NAME, block_until_done=True
         )
 
-        another_finos_waltz_users = []
```
marceloneppel, line 84: Good catch!

marceloneppel (approved):
LGTM! Thanks @TakoB222!

taurus-forever (approved):

dragomirp (approved):

PR #432, opened 2024-03-26
[DPE-3803] Show error on upgrade action

marceloneppel:
## Issue
The message that blocks an upgrade if the unit number 0 is neither a primary nor a sync-standby is shown only in the logs, making the user think that there is an issue with the cluster.

## Solution
Show the message in the action error output.


dragomirp (approved):

taurus-forever (approved):
I am afraid this message will not fit in 120 symbols juju status limit, but I believe this is your intention here (will be automatically trimmed to 120 symbols and see all details in juju debug-log).

PR #434, opened 2024-03-28
[DPE-4213] Tests legacy and modern endpoints simultaneously

BalabaDmitri:
## Issue #455 
Simultaneous creation of legacy and modern endpoints relations

## Solution
Charm goes into a "blocked" state, when creating a simultaneous relation legacy and modern endpoints.


taurus-forever:
@BalabaDmitri this is a bit over-engineered initial requirements.

The goal is to make sure the same App and DB are NOT related twice (via the modern and the legacy interfaces).
Problem: the user can forged to unrelate the legacy interface after the migration from legacy interface `pgsql` to modern interface `postgresql_client`. When keep both interfaces are connected => it could cause issues.

As discussed on sync, the test IMHO: deploy pg-test-app, deploy pg, relate app:db<>pg:db, OK, relate app:first-database pg:database, blocked, unrelated :db, OK, relate :db, blocked, unrelated :first-database, OK, unrelate :db, OK, relate :first-database OK and relation works, done. pg-test-app supports both interfaces.

We can merge this AS IS, but:
* we deploy here finos-waltz-k8s which is a bit heavy and not necessary as p-t-a is deployed anyway.
* we do not have finos-waltz-k8s to deploy on VM => we will have to simplify the test there anyway.

Merging AS IS or simplifying? Or simplifying for VM only? :-D

taurus-forever (approved):

tests/integration/relations/helpers.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python3
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+from typing import Optional
+
+import yaml
+from pytest_operator.plugin import OpsTest
+
+
+async def get_legacy_db_connection_str(
```
marceloneppel, line 10: You're not using this function in the test.

tests/integration/relations/test_relations.py
```diff
@@ -0,0 +1,175 @@
...
+    async with ops_test.fast_forward():
+        await ops_test.model.applications[APP_NAME].remove_relation(
+            f"{FINOZ_WALTZ_APP_NAME}:{DB_RELATION}", f"{APP_NAME}:{DB_RELATION}"
+        )
+        await ops_test.model.wait_for_idle(
+            apps=[FINOZ_WALTZ_APP_NAME],
+            status="blocked",
+            raise_on_blocked=False,
+            timeout=1000,
+        )
```
marceloneppel, line 155: `await ops_test.model.wait_for_idle(apps=[APP_NAME], status="active", timeout=1000)` is missing at this point.

marceloneppel (changes requested):

tests/integration/relations/test_relations.py
BalabaDmitri: @marceloneppel @taurus-forever I simplified the test

marceloneppel (approved):

taurus-forever:
I cannot merge this due to [no signed SLA](https://github.com/canonical/postgresql-k8s-operator/actions/runs/8974545200/job/25550532480?pr=434) :-(

dragomirp:
Hi, @BalabaDmitri , there's a conflict with main as well. Please, sync and resolve.

BalabaDmitri:
@dragomirp Can you check

dragomirp (approved):

PR #435, opened 2024-04-02
[DPE-3570] Add primary status message after cluster bootstrap

marceloneppel:
## Issue
The `Primary` status message is not shown when the first unit is ready to accept connections (i.e., the bootstrap of the cluster has been completed).

## Solution
Set the `Primary` status message in the primary unit right after it's ready to accept connections (while still considering the K8s primary service).

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -427,7 +427,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             self.unit_peer_data.pop("start-tls-server", None)
 
         if not self.is_blocked:
-            self.unit.status = ActiveStatus()
+            self._set_active_status()
 
     def _on_config_changed(self, event) -> None:
         """Handle configuration changes, like enabling plugins."""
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -1122,13 +1131,7 @@ def _handle_processes_failures(self) -> bool:
...
     def _set_primary_status_message(self) -> None:
         """Display 'Primary' in the unit status message if the current unit is the primary."""
-        try:
-            if self._patroni.get_primary(unit_name_pattern=True) == self.unit.name:
-                self.unit.status = ActiveStatus("Primary")
-            elif self._patroni.member_started:
-                self.unit.status = ActiveStatus()
-        except (RetryError, ConnectionError) as e:
-            logger.error(f"failed to get primary with error {e}")
+        self._set_active_status()
```
dragomirp, line 1134: We should most probably remove  this method and just call `_set_active_status()` directly.

dragomirp (approved):

src/charm.py
marceloneppel: Removing it makes total sense. Thanks for catching that. It was removed on https://github.com/canonical/postgresql-k8s-operator/pull/435/commits/569046fe7cd028adda3001393829ac4ef54d7f03.

taurus-forever (approved):
Tnx!

PR #439, opened 2024-04-04
[MISC] Update Rock and renovate tweaks

dragomirp:
* Update to latest Rock with rebuilt Patroni 3.1.2
* Update renovate config


taurus-forever:
@dragomirp is this upgrades PostgreSQL to 14.11 ? Is tests stability only the problem OR you found bigger issues? Tnx!

dragomirp:
> @dragomirp is this upgrades PostgreSQL to 14.11 ? Is tests stability only the problem OR you found bigger issues? Tnx!

IIRC we should already have 14.11 on the previous rock. I don't think there are any new issues here.

dragomirp:
> @dragomirp is this upgrades PostgreSQL to 14.11 ? Is tests stability only the problem OR you found bigger issues? Tnx!

This Rock has the new Patroni.

Issue #431, opened 2024-03-26
Enhance the error message if charm was deployed without `--trust` flag


theoctober19th:
## Steps to reproduce

1. Deploy the charm without the `--trust` flag as:
```
juju deploy postgresql-k8s --channel 14/stable
```

## Expected behavior
A helpful error message, something similar to the following would be a lot helpful:
```
Application deployed without --trust flag, which is required for this operator.
```

It looks like a similar implementation is in use in `mysql-router-k8s-operator` already:
https://github.com/canonical/mysql-router-k8s-operator/blob/105a6f6bbadaa2a5b78f8c29bab173892b5bab87/src/kubernetes_upgrade.py#L35

## Actual behavior
The charm is stuck in the following state:
```
postgresql-k8s/0*  error     idle   10.1.209.10         hook failed: "leader-elected"
```
which is a bit difficult for a layman user to debug.

## Versions

Operating system: 
Ubuntu 22.04.4 LTS

Juju CLI: 
3.4.0-genericlinux-amd64

Juju agent: 
Model  Controller  Cloud/Region        Version  SLA          Timestamp
new    k8s         microk8s/localhost  3.4.0    unsupported  08:55:30Z

Charm revision: 
193

microk8s: 
MicroK8s v1.29.2 revision 6529

## Log output

Juju debug log: 

```
unit-postgresql-k8s-0: 08:20:39 INFO juju.worker.uniter awaiting error resolution for "leader-elected" hook
unit-postgresql-k8s-0: 08:21:24 INFO juju.worker.uniter awaiting error resolution for "leader-elected" hook
unit-postgresql-k8s-0: 08:25:39 INFO juju.worker.uniter awaiting error resolution for "leader-elected" hook
unit-postgresql-k8s-0: 08:25:40 ERROR unit.postgresql-k8s/0.juju-log Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-postgresql-k8s-0/charm/venv/lightkube/core/generic_client.py", line 188, in raise_for_status
    resp.raise_for_status()
  File "/var/lib/juju/agents/unit-postgresql-k8s-0/charm/venv/httpx/_models.py", line 759, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: Client error '403 Forbidden' for url 'https://10.152.183.1/api/v1/namespaces/postgres/services/patroni-postgresql-k8s'
[...]

PR #440, opened 2024-04-08
[DPE-3885] Enhance the error message if charm was deployed without `--trust` flag

lucasgameiroborges:
## Issue

https://github.com/canonical/postgresql-k8s-operator/issues/431

## Solution

Considering this is an error that can be solved by human action directly on the live charm, I chose the following approach: If the `403 Unauthorized` error is catched, the charm enters `Blocked` status with specific message and a helpful explanation about the problem is written in the logs:

juju status:
```
Unit                         Workload  Agent      Address       Ports  Message
untrusted-postgresql-k8s/0*  blocked   executing  10.1.110.75          Insufficient permissions, try: `juju trust untrusted-postgresql-k8s --scope=cluster`
```

juju logs:
```
unit-postgresql-k8s-0: 22:26:23 INFO juju.worker.uniter.operation ran "leader-elected" hook (via hook dispatching script: dispatch)
unit-untrusted-postgresql-k8s-1: 13:24:43 ERROR unit.untrusted-postgresql-k8s/1.juju-log
            Access to k8s cluster resources is not authorized. This happens when RBAC is enabled and the deployed application was not trusted by the juju admin.
            To fix this issue, run `juju trust untrusted-postgresql-k8s --scope=cluster` (or remove & re-deploy untrusted-postgresql-k8s with `--trust`)

```


src/charm.py
```diff
@@ -625,17 +625,14 @@ def _get_hostname_from_unit(self, member: str) -> str:
...
-            self.set_secret(APP_SCOPE, REWIND_PASSWORD_KEY, new_password())
-
-        if self.get_secret(APP_SCOPE, MONITORING_PASSWORD_KEY) is None:
-            self.set_secret(APP_SCOPE, MONITORING_PASSWORD_KEY, new_password())
+        for password in {
+            USER_PASSWORD_KEY,
+            REPLICATION_PASSWORD_KEY,
+            REWIND_PASSWORD_KEY,
+            MONITORING_PASSWORD_KEY,
+        }:
```
lucasgameiroborges, line 633: this little refactor was done in order to please flake8 and reduce the amount of branching inside the function (rule `C901`)

src/charm.py
```diff
@@ -1594,6 +1604,18 @@ def get_available_resources(self) -> Tuple[int, int]:
 
         return cpu_cores, allocable_memory
 
+    def on_deployed_without_trust(self) -> None:
+        """Blocks the application and returns a specific error message for deployments made without --trust."""
+        self.unit.status = BlockedStatus(
+            "Unauthorized access to k8s resources. Is the app trusted? See logs"
+        )
```
dragomirp, line 1611: The blocked status should be removed when the application is trusted.

dragomirp:
We should add integration test that the charm will block and unblock correctly.

src/charm.py
lucasgameiroborges: During my tests, I saw that once the application gets properly trusted, a `config_changed` event gets triggered which removes the blocked status. In this case, is it necessary to explicitly unset it somewhere in the charm code?

lucasgameiroborges: I ask this because I see in the charm code that we have some cases where we set the status to blocked but I didn't see it getting explicitly unset afterwards. Maybe I missed it though.

dragomirp: As long as we don't get stuck with blocked status it should be fine. Down the road we should overhaul the way we set statuses, but that's out of scope of the PR.

lucasgameiroborges: The intended behavior we want here will be defined and verified by the integration test I'll create shortly.

tests/integration/test_trust.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_deploy_without_trust(ops_test: OpsTest):
+    """Build and deploy the charm with trust set to false.
+
+    Assert on the unit status being blocked due to lack of trust.
+    """
+    charm = await ops_test.build_charm(".")
```
dragomirp, line 91: You can use the `database_charm` fixture instead.

tests/integration/test_trust.py
```diff
@@ -0,0 +1,126 @@
...
+async def test_trust_blocked_deployment(ops_test: OpsTest):
+    """Trust existing blocked deployment.
+
+    Assert on the application status recovering to active.
+    """
+    await ops_test.juju("trust", APP_NAME, "--scope=cluster")
+
+    app = ops_test.model.applications[APP_NAME]
+    await ops_test.model.block_until(lambda: app.status in ("active", "error"), timeout=1000)
+    assert app.status == "active"
```
dragomirp, line 126:
```suggestion
    await ops_test.model.wait_for_idle(apps=[APP_NAME], status="active", timeout=1000)
```
We should wait for the entire cluster to idle. If you check the juju status [job](https://github.com/canonical/postgresql-k8s-operator/actions/runs/8638064434/job/23687044377?pr=440#step:24:13) in the CI, the replica units are still in `unknown` status

src/charm.py
```diff
@@ -1594,6 +1604,18 @@ def get_available_resources(self) -> Tuple[int, int]:
...
+        logger.error(
+            f"""
+            Access to k8s cluster resources is not authorized. This happens when RBAC is enabled and the deployed application was not trusted by the juju admin.
+            To fix this issue, run `juju trust {self._name} --scope=cluster` (or remove & re-deploy {self._name} with `--trust`)
+            """
+        )
+
```
taurus-forever, line 1618: Please update PR description accordingly. It still mention `juju resolve` there (not-aplicable for blocked state).

src/charm.py
taurus-forever, line 1611: Please consider to change Juju status to mimic `nginx-ingress-integrator`: 

```
nginx-ingress-integrator/0*  blocked      idle       10.1.204.237         Insufficient permissions, try: `juju trust <nginx-ingress-integrator> --scope=cluster`
```

P.S. IMHO, you know the app name, so you can print the fix command copy&paste'able:
```
juju deploy postgresql-k8s mypostgresql-k8s
...
...  Insufficient permissions, try: `juju trust mypostgresql-k8s --scope=cluster`
```

tests/integration/test_trust.py
taurus-forever: +1 here

src/charm.py
lucasgameiroborges: I think I had removed it, where do you see it mentioned?

lucasgameiroborges: Oh good catch, thanks!

marceloneppel (commented):
It looks great! Thanks @lucasgameiroborges!

I left only one comment.

marceloneppel (approved):

dragomirp (approved):

taurus-forever (approved):
Nice! Thank you!

PR #448, opened 2024-04-17
[DPE-2900] Cross-region async replication integration tests

marceloneppel:
## Issue
There are no integration tests for cross-region async replication.

## Solution
Add integration tests for the following scenarios:
* Set up async replication between two clusters.
* Trigger a switchover.
* Detach a cluster (turning it read-only).
* Reestablish the async replication.
* Crash the synchronous standby in the primary cluster.
* Crash the Standby Leader in the standby cluster.
* Scale the clusters.

Some of the helper functions were changed to accommodate checks on multiple models, like when checking whether writes are increasing (to now check it in both clusters).


taurus-forever (approved):
Thank you! Excellent!
P.S. please fire a followup ticket/PR for the same issue as described here: https://github.com/canonical/postgresql-operator/pull/453#pullrequestreview-2031765750

dragomirp:
Not an issue with the PR, but I don't think we'll see details about the second model in the `juju status` and `juju debug-log` CI jobs. Going forward we should think for a way to expose that via the shared workflow.

dragomirp (approved):

marceloneppel:
> Not an issue with the PR, but I don't think we'll see details about the second model in the `juju status` and `juju debug-log` CI jobs. Going forward we should think for a way to expose that via the shared workflow.

I agree. That would help us to understand what's happening in the second model when something fails.

PR #452, opened 2024-04-22
[DPE-4068] Convert all tests from unittest to pytest + enable/disable secrets in every test

lucasgameiroborges:
Follow-up to DPE-2674 in https://github.com/canonical/postgresql-operator/pull/425

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -30,20 +30,3 @@ def juju_has_secrets(mocker: MockerFixture):
             JujuVersion, "has_secrets", new_callable=PropertyMock
         ).return_value = True
         return True
-
-
-@pytest.fixture
-def only_with_juju_secrets(juju_has_secrets):
-    """Pretty way to skip Juju 3 tests."""
-    if not juju_has_secrets:
-        pytest.skip("Secrets test only applies on Juju 3.x")
-
-
-@pytest.fixture
-def only_without_juju_secrets(juju_has_secrets):
-    """Pretty way to skip Juju 2-specific tests.
-
-    Typically: to save CI time, when the same check were executed in a Juju 3-specific way already
-    """
-    if juju_has_secrets:
-        pytest.skip("Skipping legacy secrets tests")
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -3,8 +3,8 @@
 import itertools
 import json
 import logging
-import unittest
 from datetime import datetime
+from unittest import TestCase
 from unittest.mock import MagicMock, Mock, PropertyMock, patch
 
 import pytest
@@ -19,7 +19,6 @@ from ops.model import (
 )
 from ops.pebble import Change, ChangeError, ChangeID, ServiceStatus
 from ops.testing import Harness
-from parameterized import parameterized
 from requests import ConnectionError
 from tenacity import RetryError, wait_fixed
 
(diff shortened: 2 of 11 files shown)


tests/unit/conftest.py
```diff
@@ -30,20 +30,3 @@ def juju_has_secrets(mocker: MockerFixture):
...
-
-
-@pytest.fixture
-def only_without_juju_secrets(juju_has_secrets):
-    """Pretty way to skip Juju 2-specific tests.
-
-    Typically: to save CI time, when the same check were executed in a Juju 3-specific way already
-    """
-    if juju_has_secrets:
-        pytest.skip("Skipping legacy secrets tests")
```
lucasgameiroborges, line 49: We had already the whole test suite running with and without secrets (`unit-juju2` and `unit-juju3` respectively), the only change needed was to adapt the few tests that only ran in one or the other, and remove this fixtures.

tests/unit/test_charm.py
```diff
@@ -775,298 +790,282 @@ def test_validate_config_options(self, _charm_lib):
...
-            assert (
-                "Non-existing field 'non-existing-secret' was attempted to be removed"
-                in self._caplog.text
-            )
 
-    @patch("charm.JujuVersion.has_secrets", new_callable=PropertyMock, return_value=True)
-    @patch_network_get(private_address="1.1.1.1")
-    @patch("charm.PostgresqlOperatorCharm._on_leader_elected")
-    @pytest.mark.usefixtures("use_caplog")
-    def test_delete_existing_password_secrets(self, _, __):
```
lucasgameiroborges, line 950: similar to what was done in https://github.com/canonical/postgresql-operator/pull/425, this test got removed and `test_delete_password` runs with and without secrets

tests/unit/test_charm.py
```diff
@@ -775,298 +790,282 @@ def test_validate_config_options(self, _charm_lib):
...
+@pytest.mark.parametrize("scope,is_leader", [("app", True), ("unit", True), ("unit", False)])
+@patch_network_get(private_address="1.1.1.1")
+def test_migration_from_databag(harness, juju_has_secrets, scope, is_leader):
+    """Check if we're moving on to use secrets when live upgrade from databag to Secrets usage."""
+    with (
+        patch("charm.PostgresqlOperatorCharm._on_leader_elected"),
+    ):
+        # as this test checks for a migration from databag to secrets,
+        # there's no need for this test when secrets are not enabled.
+        if not juju_has_secrets:
```
lucasgameiroborges, line 987: here and on the test below, same idea from https://github.com/canonical/postgresql-operator/pull/425

tests/unit/conftest.py
lucasgameiroborges: merging those 2 suites into one will be addressed in a follow-up PR!

tests/unit/test_utils.py
```diff
@@ -2,19 +2,21 @@
...
 
-        second_password = new_password()
-        self.assertIsNotNone(re.fullmatch("[a-zA-Z0-9\b]{16}$", second_password))
-        self.assertNotEqual(second_password, first_password)
+def test_new_password():
+    # Test the password generation twice in order to check if we get different passwords and
+    # that they meet the required criteria.
+    first_password = new_password()
+    tc.assertEqual(len(first_password), 16)
+    tc.assertIsNotNone(re.fullmatch("[a-zA-Z0-9\b]{16}$", first_password))
```
dragomirp, line 18: We can just use regular `assert`s in such cases. No need to keep the test case just for that.

lucasgameiroborges: agreed, will refactor this together with merging the 2 suites in the follow-up PR.

PR #454, opened 2024-04-23
[MISC] Update data-platform-libs: data_interfaces to 34 and upgrade to 16

lucasgameiroborges:


tests/unit/test_charm.py
```diff
@@ -922,31 +920,15 @@ def test_delete_password(self, _):
...
-            assert (
-                "Non-existing field 'non-existing-secret' was attempted to be removed"
-                in self._caplog.text
-            )
-
-            self.harness.charm.remove_secret("unit", "non-existing-secret")
-            assert (
-                "Non-existing field 'non-existing-secret' was attempted to be removed"
-                in self._caplog.text
-            )
```
juditnovak, line 944: This must stay, with caplog.DEBUG. Exactly as done in https://github.com/canonical/pgbouncer-operator/pull/207/files#diff-0580ab9aef31b44580478eb270ec3422c5e6fef35df585fecb3a68317e246809L639

juditnovak (changes requested):

src/charm.py
```diff
@@ -261,6 +261,9 @@ def get_secret(self, scope: Scopes, key: str) -> Optional[str]:
             raise RuntimeError("Unknown secret scope.")
 
         peers = self.model.get_relation(PEER)
+        if not peers:
```
marceloneppel, line 264: Won't this charm need `DataPeerData` and `DataPeerUnitData`, like we see when the lib was updated in the other charm repos?

lucasgameiroborges: oh yes, in fact this PR is actually still a WIP, I forgot to change it to draft! Sorry about that

lucasgameiroborges: PR is improved now, feel free to re-review!

marceloneppel (approved):

taurus-forever (approved):

tests/unit/test_charm.py
```diff
@@ -870,20 +873,14 @@ def test_set_secret(harness):
...
         assert "password" not in harness.get_relation_data(rel_id, harness.charm.app.name)
 
         # Test unit scope.
         assert "password" not in harness.get_relation_data(rel_id, harness.charm.unit.name)
         harness.charm.set_secret("unit", "password", "test-password")
-        assert (
-            harness.get_relation_data(rel_id, harness.charm.unit.name)["password"]
-            == "test-password"
-        )
+        assert harness.charm.get_secret("unit", "password") == "test-password"
```
juditnovak, line 883: We shouldn't re-write the old test.

We should keep Juju2 compatibilty.

Please add a separate test for secrets, that's only executed when secrets are available

juditnovak (changes requested):

tests/unit/test_charm.py
lucasgameiroborges: I kept the old test, running on juju2 only, as it checks the relation databag for the secret. I kept the changed test running on both versions as it is compatible either way. Left some comments for future perspective. LMK what you think

juditnovak: All right, I see, sorry it was my confusion.

Now I understand what you mean, renaming was a good idea here :-)

All good on this one, thank you :-)

tests/unit/test_charm.py
```diff
@@ -862,7 +865,11 @@ def test_get_secret_secrets(harness, scope):
 
 
 @patch_network_get(private_address="1.1.1.1")
-def test_set_secret(harness):
+def test_set_secret_in_databag(harness, juju_has_secrets):
+    # As this test set secrets using set_secret and checks if they end up
+    # inside the relation databag, this is meant to run on juju2 only
+    if juju_has_secrets:
+        return
```
juditnovak, line 872: We shouldn't have test returning on a condition.

`pytest.makr.skip` was invented for this reason :-)

Could you pls use something like my fixtures instead? 
https://github.com/canonical/data-platform-libs/blob/main/tests/conftest.py#L37-L51

juditnovak (changes requested):

tests/unit/test_charm.py
juditnovak: Actually... My bad sorry, I overlooked this test.

This one is not using secrtes.

IT's only using secrets when secrets are configured but otherwise it's falling back to the databag.

We should have a test that is specifically checking if a Juju secret comes to existence when we create one with `charm.set_secret()`.

I think we had it already somewhere lower down in this file.

We don't need to duplicate that.

I mean: it's good to have a single generic test on set-get secret functionality. It will work both on databag and with Juju Secrets, with no knowledge of the underlying storage.

However we also MUST have specific tests for both databag usage (`@only_without_juju_secrets` -- should fail otherwise) and vica versa.

I think I have some of those in all PG apps `main` right now. That concept has to be kept and followed if we want a single pipeline.

dragomirp (approved):

tests/unit/test_charm.py
```diff
@@ -204,6 +204,7 @@ def test_on_postgresql_pebble_ready_no_connection(harness):
 
 def test_on_get_password(harness):
     # Create a mock event and set passwords in peer relation data.
+    harness.set_leader(True)
```
juditnovak, line 207: All units should be able to get the password. If now we reduce it to the leader, we're changing the expected behavior instead of the previously expected behavior of the application.

I mean: earlier this test was ensuring that all units get the password (in a way as described by the test).
Setting the unit to `leader`, from now on this test only ensures (expect) that we can get the password this way for the leader. 

Tests generally _mustn't_ change -- except when being aligned to a new behavior (i.e. fix or feature) we are implemneting. However, the changes we did in the charm should have been fully identical to the previous behavior. 

There was supposed to be no behavior change in this PR -- thus the tests have to stay intact, too.

Was there a reason for switching to leader? I may be overlooking something, perhaps we did a change that stopped the test from passing as before (and we may want to investigate that).

lucasgameiroborges: Yes, this specific test was failing after the library bump.

My reason to add it was: I realized that this exact same test on the postgreSQL VM repo had `set_leader(True)` in it ([this commit,](https://github.com/canonical/postgresql-operator/commit/368dba73450d8a640bb1b1847d79454c1e639055) we can see it being added in the diff). At first I saw no reason for this test to be different on k8s and vm, so I added `set_leader(True)` here in order to make them coherent with each other, and this change made it pass again.

I thought this was a fair addition to the test, since the same exact test over on VM had it in the first place (not sure why we didn't have it here). If there was a reason for it to not be here, then I might need some more context and suggestions on what to do about it (if not having this).

I think its a good idea to have @marceloneppel and @dragomirp input here, as they might have context that I lack.

dragomirp: @juditnovak I think the issue is that `fetch_my_relation_field` is marked as leader only:
https://github.com/canonical/postgresql-k8s-operator/blob/lucas/update-lib/lib/charms/data_platform_libs/v0/data_interfaces.py#L1543

dragomirp: I'd propose merging this as is, until we establish the correct behaviour of the translation logic, since we already have it everywhere else.

juditnovak: Sure @dragomirp if U prefer, go head.

lucasgameiroborges:
Following our discussion about why `test_on_get_password` is passing only if the unit is leader, I created the following ticket to address it: https://warthogs.atlassian.net/browse/DPE-4259
If everyone agrees I think this PR can get merged in the meantime. Then I can create follow-ups once we investigate the issue.

marceloneppel:
I think that's fine to merge it as is and investigate later.

PR #466, opened 2024-05-02
[MISC] Always check peer data for legacy secrets

dragomirp:
Follow up to https://github.com/canonical/postgresql-k8s-operator/pull/454

Try to always do the translation from peer data to secrets in case of very old revisions being updated.


juditnovak (approved):
So, normally this is somethig that the lib was supposed to do for you.

However (thanks to the wonderful Postgres unittests -- that now will be also migrated to the lib test coverage) I had to realize that there is a bug on this behavior with the new `get_secret()`, `set_secret()` methods.

https://github.com/canonical/data-platform-libs/issues/174

So please go ahead with the change (for now).

PR #474, opened 2024-05-10
Add Differential backups support

delgod:


taurus-forever:
Aborted as no longer topical, see https://github.com/canonical/postgresql-k8s-operator/pull/476

PR #481, opened 2024-05-20
[DPE-4336] Reset active status when removing extensions dependency block

dragomirp:
Port of https://github.com/canonical/postgresql-operator/pull/467

* The charm should not restore the blocked status if the blocking condition is no longer valid
* Skip delete pod test in backups if there are no secrets

### DIFF
--- a/tests/integration/test_backups.py
+++ b/tests/integration/test_backups.py
@@ -430,7 +430,7 @@ async def test_invalid_config_and_recovery_after_fixing_it(
 
 
 @pytest.mark.group(1)
-async def test_delete_pod(ops_test: OpsTest):
+async def test_delete_pod(ops_test: OpsTest, github_secrets) -> None:
     logger.info("Getting original backup config")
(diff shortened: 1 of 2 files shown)


tests/integration/test_backups.py
```diff
@@ -430,7 +430,7 @@ async def test_invalid_config_and_recovery_after_fixing_it(
 
 
 @pytest.mark.group(1)
-async def test_delete_pod(ops_test: OpsTest):
+async def test_delete_pod(ops_test: OpsTest, github_secrets) -> None:
```
dragomirp, line 433: The test depend on the preceding tests, so it should also be skipped if there are no secrets.

lucasgameiroborges (approved):

marceloneppel (approved):
LGTM!

taurus-forever (approved):
Merging PR as all LGTM here and Dragomir is on vacation. Tnx!

PR #482, opened 2024-05-22
chore: fix rocks terminology

cjdcordeiro:
## Issue
"ROCK" should be "rock"

## Solution

Replace all occurrences


marceloneppel (approved):

taurus-forever (approved):
TIL. Sad. ROCK is ROCK, and rock is almost pebble. :-D

PR #476, opened 2024-05-15
[DPE-4408] Add differential Backup support

lucasgameiroborges:
## Issue
Link to issue: https://warthogs.atlassian.net/browse/DPE-4408
Add support for differential backup, plus a new parameter `type` to `create-backup` action for selecting the backup type (`full` or `diff`).

## Solution

- Implement new param in `create-backup` juju action.
- Include the backup type in `list-backups` action output.
- Refactor to make backup charm logic handle different backup types (label formatting).
- Adapt integration test to try both full and differential backups.
- Small fine-tuning in integration and unit tests.

## Follow-ups

- Update documentation to mention differential backups: https://warthogs.atlassian.net/browse/DPE-4357
- Investigate the current main source of instability in integration test: https://warthogs.atlassian.net/browse/DPE-4358
- Replicate this PR to PG VM: https://warthogs.atlassian.net/browse/DPE-4389

PR #487, opened 2024-05-28
[DPE-4464] Add Incremental backup support + proper type naming

lucasgameiroborges:
## Issue

- Adds support for incremental type backups
- Proper naming usage (in accordance to [spec](https://docs.google.com/document/d/1WWMt_Q71cUIrpfYirp5TRCdkJXMjFSWCH6CTnKrO4YA/edit))

## Solution

Follow-up to #476 

## Follow-up

https://warthogs.atlassian.net/browse/DPE-4475 for including incremental backup check in integration test.

### DIFF
--- a/src/backups.py
+++ b/src/backups.py
@@ -296,13 +296,13 @@ class PostgreSQLBackups(Object):
         """Parse backup ID as a timestamp."""
         if label[-1] == "F":
             timestamp = label
-            backup_type = "FULL"
+            backup_type = "full"
         elif label[-1] == "D":
             timestamp = label.split("_")[1]
-            backup_type = "DIFFERENTIAL"
+            backup_type = "differential"
         elif label[-1] == "I":
             timestamp = label.split("_")[1]
-            backup_type = "INCREMENTAL"
+            backup_type = "incremental"
         else:
             raise ValueError("Unknown label format for backup ID: %s", label)
 
@@ -474,7 +474,7 @@ class PostgreSQLBackups(Object):
 
     def _on_create_backup_action(self, event) -> None:  # noqa: C901
         """Request that pgBackRest creates a backup."""
-        backup_type = event.params.get("type", "FULL")
+        backup_type = event.params.get("type", "full")
         if backup_type not in BACKUP_TYPE_OVERRIDES:
             error_message = f"Invalid backup type: {backup_type}. Possible values: {', '.join(BACKUP_TYPE_OVERRIDES.keys())}."
             logger.error(f"Backup failed: {error_message}")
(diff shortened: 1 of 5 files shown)


delgod (approved):

src/backups.py
```diff
@@ -474,16 +474,14 @@ def _on_s3_credential_changed(self, event: CredentialsChangedEvent):
...
-            )
+        backup_type = event.params.get("type", "full")
+        if backup_type not in BACKUP_TYPE_OVERRIDES:
+            error_message = f"Invalid backup type: {backup_type}. Possible values: {', '.join(BACKUP_TYPE_OVERRIDES.keys())}."
             logger.error(f"Backup failed: {error_message}")
             event.fail(error_message)
             return
 
-        logger.info(f"A {backup_type} backup  has been requested on unit")
+        logger.info(f"A {backup_type} backup has been requested on unit")
```
marceloneppel, line 484: Is it expected to create a full backup if there is no previous backup and the user requests an incremental backup?

tests/integration/test_backups.py
```diff
@@ -170,7 +170,7 @@ async def test_backup_and_restore(ops_test: OpsTest, cloud_configs: Tuple[Dict,
         # Run the "create backup" action.
         logger.info("creating a backup")
         action = await ops_test.model.units.get(replica).run_action(
-            "create-backup", **{"type": "diff"}
+            "create-backup", **{"type": "differential"}
```
marceloneppel, line 173: A small test for incremental backup would be important.

lucasgameiroborges: Though I agree, I fear adding it now (i.e. introducing 2 extra heavy operations, incremental backup creation + restore) will make the integration test even more long and unstable due to scaling/wait_for_idle (this happened with differential backups before, causing [this story](https://chat.canonical.com/canonical/pl/qzq5cfq7i3gtbrcmhqg4pobtmy) ) That's why I created https://warthogs.atlassian.net/browse/DPE-4475 as a follow-up: Once the test has improved in terms of stability and/or speed, then we could add it without risk of degrading the CI. LMK what you think.

marceloneppel: My bad. I forgot about this follow-up you created.

src/backups.py
lucasgameiroborges: AFAIU this behavior comes from pgbackrest: It checks internally if it's possible to create a differential/incremental backup; if it's not possible (for any reason), it creates a full backup.
I suppose we could implement this check in the charm ourselves and return error to the user if not possible, but IMO this wouldn't be an easy check to get right, so that's why I left it up to pgbackrest. LMK what's the preferred approach here.

taurus-forever: Let's follow the default pgbackrest behaviour here. K.I.S.S.
Please re-confirm it with Mohamed in spec.

taurus-forever (approved):

src/backups.py
lucasgameiroborges: If Mohamed says we should do things differently regarding this behavior, it will probably deserve a separate Jira ticket/PR for the change, as it would be out of scope of this one (which adds incremental option)

marceloneppel (approved):

lucasgameiroborges:
I'll go ahead and merge this feature, then. Regarding [this raised issue](https://github.com/canonical/postgresql-k8s-operator/pull/487#discussion_r1619258356), I'll create a dedicated ticket/PR for it in case we want to proceed differently from the default behavior, since it would be out-of-scope of this PR. Thanks!

PR #489, opened 2024-05-28
[DPE-4456] Fix scale up with S3 and TLS relations

marceloneppel:
## Issue
We cannot scale up if S3 and TLS relations are established. This operation leads to an error like the one below if, in the new unit, the S3 parameters relation changed event fires before the TLS files are available in the unit (and Patroni + PostgreSQL haven't started yet).
```sh
requests.exceptions.SSLError: HTTPSConnectionPool(host='10.142.152.2', port=8008): Max retries exceeded with url: /cluster (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1007)')))
```

## Solution
Defer the S3 parameters relation changed event until that unit can do any operation related to S3 (or even connect to the other unit to check the cluster topology and verify it's not the primary unit so that it can abort any kind of stanza initialisation).

Port of https://github.com/canonical/postgresql-operator/pull/480.

### DIFF
--- a/src/backups.py
+++ b/src/backups.py
@@ -82,6 +82,22 @@ class PostgreSQLBackups(Object):
 
         return True, None
 
+    @property
+    def _can_initialise_stanza(self) -> bool:
+        """Validates whether this unit can initialise a stanza."""
+        # Don't allow stanza initialisation if this unit hasn't started the database
+        # yet and either hasn't joined the peer relation yet or hasn't configured TLS
+        # yet while other unit already has TLS enabled.
+        if not self.charm._patroni.member_started and (
+            (len(self.charm._peers.data.keys()) == 2)
+            or (
+                "tls" not in self.charm.unit_peer_data
+                and any("tls" in unit_data for _, unit_data in self.charm._peers.data.items())
+            )
+        ):
+            return False
+        return True
+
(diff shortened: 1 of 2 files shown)


src/backups.py
```diff
@@ -455,6 +471,11 @@ def _on_s3_credential_changed(self, event: CredentialsChangedEvent):
             logger.debug("Cannot set pgBackRest configurations, missing configurations.")
             return
 
+        if not self._can_initialise_stanza:
+            logger.debug("Cannot initialise stanza yet.")
```
lucasgameiroborges, line 475: Should this be DEBUG level only? Kinda feel a little more important than this to me, but up to you!

lucasgameiroborges (approved):
LGTM! Left only one question.

taurus-forever (approved):
LGTM, but 30 mins timeout scary me a lot...

src/backups.py
marceloneppel: I think DEBUG is okay for this case. Like the other messages we already have for similar situations, which don't provide immediate value for the user but only when troubleshooting, I think it's okay as it is.

PR #496, opened 2024-06-04
[DPE-4427] Address main instability sources on backups integration tests

lucasgameiroborges:
## Issue

The `test_backups.py` integration test has been one of the main sources of CI failure in recent past. This PR aims to help stabilize the test.

## Solution

This PR addresses a number of root causes to the problem, namely:

- [DPE-4427] `hook failed: "certificates-relation-changed"`: The function call `push_tls_files_to_workload()` was failing with a transient connection error after multiple retries, but tenacity's `RetryError` was not one of the catched exceptions, thus making the hook fail instead of deferring.

- [DPE-4425] charm stuck on scale-up: Caused by an infinite defer loop on `on_peer_relation_changed` event because cluster was not initialized. Added an extra check to initialize cluster if on leader unit, which breaks the loop.

- [DPE-2107] Switchover on scale-down: If the primary unit is the one being removed in a scale-down, the charm doesn't recover ==> implement a switchover on `storage_detaching` hook.

- [No ticket?] Add connection check after restart operation: implements a `_can_connect_to_postgresql` check that is verified on restart hook.

- [No ticket?] `update_read_only_endpoint` call may fail due to permission denied when updating databag/secret in the case of read-only cluster in the async_replication test ==> catch and log the error and move forward with the event.

## Related Follow-ups

- `test_backups.py` charm may get stuck when TLS and S3-config events happen at the same time (?): https://warthogs.atlassian.net/browse/DPE-4596
- `test_charm.py` instability over the `shared_buffers` setting: https://warthogs.atlassian.net/browse/DPE-4594
- `test_tls.py` instability due to `pg_rewind` sometimes not using TLS: https://warthogs.atlassian.net/browse/DPE-4590

If you see an immediate solution for the issues mentioned, please let me know so I can test it and include in the PR. Otherwise, we have tickets for further investigation.

[DPE-4427]: https://warthogs.atlassian.net/browse/DPE-4427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-4425]: https://warthogs.atlassian.net/browse/DPE-4425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DPE-2107]: https://warthogs.atlassian.net/browse/DPE-2107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/integration/test_backups.py
```diff
@@ -142,7 +147,9 @@ async def test_backup_and_restore(ops_test: OpsTest, cloud_configs: Tuple[Dict,
             )
         connection.close()
 
-        # Run the "create backup" action.
+        # With a stable cluster, Run the "create backup" action
+        async with ops_test.fast_forward():
+            await ops_test.model.wait_for_idle(status="active", timeout=1000, idle_period=30)
```
lucasgameiroborges, line 152: This was added because, in rare cases, the charm was not idle/ready for a backup to be created.

src/charm.py
```diff
@@ -437,7 +493,10 @@ def _on_peer_relation_changed(self, event: HookEvent) -> None:
             event.defer()
             return
 
-        self.postgresql_client_relation.update_read_only_endpoint()
+        try:
+            self.postgresql_client_relation.update_read_only_endpoint()
+        except ModelError as e:
+            logger.debug("Cannot update read_only endpoints: %s", str(e))
```
dragomirp, line 499: This should most probably be logged as error.

lucasgameiroborges: I thought about this, but AFAICT this error occurring (and getting caught here) did not immediately affected the charm's behavior, and the user wouldn't be able to do anything about it (charm doesn't get blocked or anything), so would it be useful to have it as a big red error log? @marceloneppel what do you think?

I don't have a strong opinion on this, just trying to better understand what it takes for a log to be high level (warning/error)

tests/integration/test_charm.py
```diff
@@ -174,7 +174,7 @@ async def test_postgresql_parameters_change(ops_test: OpsTest) -> None:
     """Test that's possible to change PostgreSQL parameters."""
     await ops_test.model.applications[APP_NAME].set_config({
         "memory_max_prepared_transactions": "100",
-        "memory_shared_buffers": "128",
+        "memory_shared_buffers": "32768",  # 2 * 128MB. Patroni may refuse the config if < 128MB
```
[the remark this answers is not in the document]
lucasgameiroborges: You mean converted from (# of blocks of 8Kb) to Mb? The value I set is 256Mb in # of 8Kb blocks, which is the appropriate unit [as per here](https://github.com/canonical/postgresql-k8s-operator/blob/8f7ec66d533b0909f6f712715f379f741e89f661/config.yaml#L62) no? The  previous value of `128 ==> 1024 Kb` was too small I think?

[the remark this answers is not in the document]
lucasgameiroborges: You linked the conversion of the maximum threshold (from Mb to blocks of 8Kb), which should be much bigger than 256 Mb

taurus-forever (approved):
Let me trust CI/CD here. LGTM! Thank you!

lib/charms/postgresql_k8s/v0/postgresql_tls.py
```diff
@@ -142,7 +143,7 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
                 logger.debug("Cannot push TLS certificates at this moment")
                 event.defer()
                 return
-        except (ConnectionError, PathError, ProtocolError) as e:
+        except (ConnectionError, PathError, ProtocolError, RetryError) as e:
```
marceloneppel, line 146: You also need to bump the `LIBPATCH` to 9.

src/charm.py
marceloneppel: A warning is better as it didn't affect the charm behavior, but may eventually lead to undesirable effects later.

tests/integration/test_charm.py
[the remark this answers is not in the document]
marceloneppel: It seems that we're passing the wrong value in the test. Let's double-check it in the ticket related to the shared buffers value increase/decrease.

marceloneppel (approved):
LGTM! Please just update the PostgreSQL TLS library `LIBPATCH`.

src/charm.py
lucasgameiroborges: Warning it is!

tests/integration/test_charm.py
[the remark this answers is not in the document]
lucasgameiroborges: Ticket in question here: https://warthogs.atlassian.net/browse/DPE-4594

marceloneppel (approved):

dragomirp (approved):

PR #502, opened 2024-06-10
[MISC] Update microk8s to 1.30

dragomirp:


dragomirp:
Should be handled by renovate

PR #510, opened 2024-06-14
[DPE-4594][DPE-4620] Address test_charm and test_self_healing instabilities

lucasgameiroborges:
## Issue

Address some more of the instabilities on the postgresql-k8s CI. 

## Solution

For [DPE-4594]: Two things, 
 - There was some confusion (to me, at least!) about the parameter being in Mb or in number of blocks of 8Kb . AFAICT the proper unit, according to both patroni/postgres and our config, is blocks of 8Kb and thus I made it so everywhere in the code (conversion being 1 Mb = 128 blocks of 8Kb).
  - Shared memory configs should go to patroni too ==> `shared_buffers` added to `bulk_update_parameters_controller_by_patroni`

For [DPE-4620]: Not much had to change here, just adding some proper retries/error-catching logic to some error-prone steps like `test_network_cut` function calls greatly improved the duration/stability of the test.

[DPE-4594]: https://warthogs.atlassian.net/browse/DPE-4594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-4620]: https://warthogs.atlassian.net/browse/DPE-4620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/patroni.py
```diff
@@ -306,13 +306,30 @@ def member_started(self) -> bool:
             allow server time to start up.
         """
         try:
-            for attempt in Retrying(stop=stop_after_delay(90), wait=wait_fixed(3)):
+            for attempt in Retrying(stop=stop_after_delay(30), wait=wait_fixed(3)):
```
lucasgameiroborges, line 309: This whole change to the `member_started` check is more a suggestion, I think it could be useful to check with inverted `TLS_enabled` flag in case it fails (also think this improves stability). Apparently this change made codecov sad because we don't have test coverage for this function? LMK what you think about it

tests/integration/ha_tests/helpers.py
```diff
@@ -481,23 +483,28 @@ async def get_sync_standby(model: Model, application_name: str) -> str:
...
-    password = await get_password(ops_test, database_app_name=app, unit_name=unit_name)
-    address = await get_unit_address(ops_test, unit_name)
     try:
+        app = unit_name.split("/")[0]
         for attempt in Retrying(stop=stop_after_delay(60), wait=wait_fixed(3)):
             with attempt:
+                password = await asyncio.wait_for(
+                    get_password(ops_test, database_app_name=app), 15
+                )
+                address = await asyncio.wait_for(get_unit_address(ops_test, unit_name), 15)
```
lucasgameiroborges, line 495: Important to have timeouts for those calls, they were hanging indefinitely in the network cut test

lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -604,12 +604,11 @@ def build_postgresql_parameters(
                 # Use 25% of the available memory for shared_buffers.
                 # and the remaining as cache memory.
                 shared_buffers = int(available_memory * 0.25)
+                parameters["shared_buffers"] = f"{int(shared_buffers * 128 / 10**6)}"
             effective_cache_size = int(available_memory - shared_buffers)
-            parameters.setdefault("shared_buffers", f"{int(shared_buffers / 10**6)}MB")
-            parameters.update({"effective_cache_size": f"{int(effective_cache_size / 10**6)}MB"})
-        else:
-            # Return default
-            parameters.setdefault("shared_buffers", "128MB")
```
lucasgameiroborges, line 612: I removed this `setdefault` of 128MB because that's what postgresql have as a default normally? No need to set it like this AFAIK, LMK if I'm missing something

src/patroni.py
```diff
@@ -306,13 +306,30 @@ def member_started(self) -> bool:
...
+                self._tls_enabled = not self._tls_enabled
+                return False
 
-        return r.json()["state"] in RUNNING_STATES
+        if r.json()["state"] in RUNNING_STATES:
+            logger.debug("Patroni health check succeeded.")
+            return True
+        else:
+            logger.debug(f'Patroni online, but unit in bad state: {r.json()["state"]}')
+            return False
```
taurus-forever, line 332: I have noticed, the patroni often replies 500 the very first request, and then 200 all the following requests the next second... and we reject/defer the whole event processing due to those one 500. Maybe we can save out time here to accept the very first 500 once OR at least try to improve this upstream and have no 500 from patroni on boot.

taurus-forever (approved):
Up 2 @marceloneppel here.
Code is OK, but cannot comment charges.

src/patroni.py
lucasgameiroborges: We normally retry the call at least 10 times here before giving up, no? When you say the first request, you mean this first batch of requests+retries? For me it seems a little risky to accept 500 as a valid response, probably worth trying to understand this better. Have any examples on CI/ticket for it? I'll keep an eye

lib/charms/postgresql_k8s/v0/postgresql.py
marceloneppel: You're right. I tested with the `testing` profile, and it worked correctly, even after changing the configuration and later resetting it.

src/patroni.py
marceloneppel: Did you track down the specific situation where the inverted `TLS_enabled` flag was needed? I mean, in a very specific situation, we are probably setting the wrong value to `self._tls_enabled`. IMO, that information is really important.

marceloneppel (commented):
The part about `shared buffers` LGTM! I'm curious about the specific case that led to the need for the inverted `TLS_enabled` flag.

src/patroni.py
dragomirp: Maybe we can also check if the cert files exist in `self._tls_enabled`? I can't think of another reason for it to be incorrect.

lucasgameiroborges: After further testing, I decided to revert this specific change. During my initial testing, I was led to believe that this extra check was improving test stability, but now after re-analysis I don't see any clear indication that it improved anything. Therefore, for the time being, I think its better to not include it.

I'll keep an eye out to see if any issues arise due to this check failing because of wrong `tls_enabled`, but that will be for outside this PR. Thanks for the discussion!

dragomirp (approved):

src/patroni.py
taurus-forever: Here is the pricese example of my story in 14/edge: https://pastebin.canonical.com/p/HXGpNqnsNJ/

As you can see we reloaded patroni and hit 503 in line 25 => deffered => timeout for deployment until the juju res-send event. Unnecessary waiting.... :sob:

marceloneppel (approved):

src/patroni.py
lucasgameiroborges: I created the following ticket with the information we have on this at the moment: https://warthogs.atlassian.net/browse/DPE-4728
We should probably consider some alternatives, I may have some ideas on this, will keep an eye!

PR #524, opened 2024-06-21
[DPE-4620][MISC] CI stability fixes + slicing tests

lucasgameiroborges:
This PR encompasses a number of medium-to-small changes that improve stability on the integration tests.

Some of the changes include:
 - Adds `check_freq=3` to reduce check frequency in `wait_for_idle` calls to 3 seconds (instead of 0.5 seconds default). This should reduce the chance of getting transient  `JujuAppError` type errors from `python-libjuju` when deploying the charm, though this problem still exists and affects some of our tests randomly, will create a ticket for it.
 - Reduce retry intervals for some patroni health checks, which were too long and just causing delays to inevitable defers (and causing timeouts).
  - `test_scaling_to_zero` in self healing test got separated in a second group.
  - `test_kill_db_process` and `test_terminate_db_process` got merged into one parametrized `test_interruption_db_process` with slight changes to improve stability
   - `_connect_to_database` fuction in lib now has retry logic to avoid transient errors.
   
   ## Follow-ups
   Main pain points on the CI stability after merge:
   https://warthogs.atlassian.net/browse/DPE-4590 for `test_tls`,
   https://warthogs.atlassian.net/browse/DPE-4854 for `test_self_healing`,
   https://warthogs.atlassian.net/browse/DPE-4855 for all tests,
   (Among other, less frequent issues that still arise in the CI, just don´t have a ticket for them yet)
   
   And https://warthogs.atlassian.net/browse/DPE-4856 for porting the fixes to VM charm.


lucasgameiroborges:
Last commit CI run: green on 2nd try
Second to last commit CI run: green on 3rd try

Looking good 😃

tests/integration/helpers.py
```diff
@@ -101,6 +104,7 @@ async def build_and_deploy(
             status=status,
             raise_on_blocked=True,
             timeout=1000,
+            check_freq=3,
```
dragomirp, line 107: I think we should set `raise_on_error=False` here instead of changing the frequency everywhere. If the volume error is transient and happens only on deploy, we should only try to suppress it on deploy.

lucasgameiroborges: But it doesn't happen on deploy exactly, it happens on wait_for_idle (and we probably don't want to disable raising on error completely on there)

lucasgameiroborges: It is on an early stage on the charm deployment (during agent install) but it's the wait_on_idle call who raises the error

dragomirp: `build_and_deploy()` will deploy and `wait_for_idle()`. If the error is happening on the wait after the deploy, that's where we should guard for the error. The worse case scenario for setting `raise_on_error=False` only during `build_and_deploy()` is that we will have to wait the full timeout for the deploy to fail, since if the error doesn't heal, `wait_for_idle()` should still fail in the end. Tweaking `check_freq` everywhere, on the other hand, could also hide other issues beyond the volume mount error.

lucasgameiroborges: I'm not sure I get your point: the way I see it, increasing check frequency would only "hide" errors which are transient and resolve in < 3s, which to me is fine. Avoiding raise on errors, on the other hand, will not only hide the same transient errors, but will hide non-transient errors as well, and we'll only see what happened after timeout. I personally don't see the benefit.

lucasgameiroborges: Testing Try-catch with retry approach here https://github.com/canonical/postgresql-k8s-operator/pull/557

tests/integration/test_tls.py
```diff
@@ -166,6 +166,11 @@ async def test_mattermost_db(ops_test: OpsTest) -> None:
             with attempt:
                 await check_tls_rewind(ops_test)
 
+        # Await for postgresql to be stable if not already
+        await ops_test.model.wait_for_idle(
+            apps=[DATABASE_APP_NAME], status="active", idle_period=15
+        )
+
```
marceloneppel, line 173: Is the charm not being completely ready the reason the test fails? If so, we should fix it differently.

We should figure out if it is related to the connection string being shared already in the relation data, but the database is not ready yet, or if it's for another reason that the failure happens.

This can be addressed in a separate PR.

lucasgameiroborges: The motivation for this was to avoid errors like this one: https://github.com/canonical/postgresql-k8s-operator/actions/runs/9813699722/job/27100493808#step:24:652 where the charm fails to initialize the `db` relation. 

AFAICT this happens because we did not allow enough time for the charm to recover after running `await run_command_on_unit(ops_test, primary, "/charm/bin/pebble start postgresql")` so thats why I added this wait right after. 

Is the charm expected to recover immediately after this sort of restart (running `pebble stop/start` directly)? If so, then I agree, this is a bug and we should probably address the underlying issue in another PR. If not, then IMO its okay to keep this wait. Up to you guys

lucasgameiroborges: Oh, actually I think I got your point: Even if the charm is not immediately ready after the restart, adding the relation should not fail, but instead delay the sharing of the connection string. Is this what you mean?

marceloneppel: Exactly.

However, after reading your comment about `pebble start` and thinking better, I think it's ok to leave it as is for now because we are doing some operations manually in the test (like the `pebble start`).

It would be good if the charm detects this situation (that the pebble service is offline) and removes the instance from the list of ready units (in the services used for the connection strings), but it would require more work. You can create a ticket so we can remember about that. Thanks!

marceloneppel (approved):

tests/integration/helpers.py
lucasgameiroborges: Try-catching the error didn´t work so well, since it requires some effort to effectively retry the deployment on `build_and_deploy` helper, and still in a lot tests we don't even use this function to deploy (some tests where we deploy and relate 2 or more charms at the same time).

After some test and consideration, I think Dragomir's suggestion is indeed the best approach here: Adding `raise_on_error=False` to `build_and_deploy`'s `wait_for_idle' call (and, in the case of tests who don´t use it, to the first `wait_for_idle` call right after deploy) is a sure way to avoid the bug in the specific place where it happens + If anything goes wrong later in the execution, by then we'll have logs to help debug the problem.

I'll adapt this PR to remove all the `check_freq`s and use `raise_on_error=False` in the places where I think it makes sense. LMK what you think.

tests/integration/test_tls.py
lucasgameiroborges: Created https://warthogs.atlassian.net/browse/DPE-4866 for future investigate this question. For the time being, I'll keep this then, to avoid CI suffering.

tests/integration/helpers.py
lucasgameiroborges: Huh, it ended up a bit ugly since a lot of places either don't use `build_and_deploy`, or use it with `wait_for_idle=False` and waiting afterwards... LMK if you agree with the changes or see a better way.

tests/integration/helpers.py
```diff
@@ -304,6 +308,7 @@ async def deploy_and_relate_application_with_postgresql(
         apps=[application_name],
         status="active",
         raise_on_blocked=False,  # Application that needs a relation is blocked initially.
+        raise_on_error=False,
```
lucasgameiroborges, line 311: This is to avoid transient errors after deploy for charms that we relate to postgres using this helper function. This helped avoid some of the errors in `mattermost` charm on TLS test, like I commented today.

dragomirp (approved):

lucasgameiroborges:
I'll go ahead and merge this, I think discussions were either addressed or have a ticket associated to it. Thank you!

PR #528, opened 2024-06-25
[MISC] Suppress oversee user

dragomirp:


dragomirp:
We don't seem to oversee users, so there's no point in adding this yet.

PR #532, opened 2024-06-26
[DPE-1177] Change 'master' to 'primary' in Patroni leader role

marceloneppel:
## Issue
Patroni sets some labels in the pods to indicate which unit is the primary and which are the replicas. Unfortunately, in older versions, it's using the `master` value for the `role` label.

## Solution
We upgraded to use the `primary` value for the `role` label because we wanted to use better language. It's now possible to use custom values for the labels.

Notes:
* The version constraints were upgraded on `src/dependency.json` to later, in the next stable after the one that will contain these changes, we can remove the upgrade logic by updating the version constraints again.
* `src/charm.py` don't patch the charm custom k8s services during an upgrade to avoid creating moments of downtime (like, for example, if there is a unit still with a role label equal to `master` and we already switch to the value equal to `primary` in the k8s service).
* `src/upgrade.py` contains temporary logic to handle the label value update during the upgrade process (moving to the `primary` label only when the last unit - unit 0 - will be upgraded to avoid downtime).
* `tests/integration/ha_tests/test_rollback_to_master_label.py` and `tests/integration/ha_tests/test_upgrade_to_primary_label.py` are almost a copy from `tests/integration/ha_tests/test_upgrade.py`, but they should be removed later, as the intent is only to validate the role label value migration at this point.
* More unit tests will be added through another PR because this PR is already over 500 lines.


tests/integration/ha_tests/test_rollback_to_master_label.py
```diff
@@ -0,0 +1,174 @@
...
+@pytest.mark.group(1)
+@markers.amd64_only  # TODO: remove after arm64 stable release
+@pytest.mark.abort_on_fail
+async def test_deploy_stable(ops_test: OpsTest) -> None:
+    """Simple test to ensure that the PostgreSQL and application charms get deployed."""
+    await asyncio.gather(
+        ops_test.model.deploy(
+            DATABASE_APP_NAME,
+            num_units=3,
+            channel="14/stable",
```
dragomirp, line 44: Maybe we should use a specific revision here? New releases would use the new label by default.

delgod (approved):

tests/integration/ha_tests/test_rollback_to_master_label.py
```diff
@@ -0,0 +1,174 @@
...
+        ops_test.model.deploy(
+            APPLICATION_NAME,
+            num_units=1,
+            channel="latest/edge",
+        ),
+    )
+    logger.info("Wait for applications to become active")
+    async with ops_test.fast_forward():
+        await ops_test.model.wait_for_idle(
+            apps=[DATABASE_APP_NAME, APPLICATION_NAME], status="active"
```
lucasgameiroborges, line 56: Optional of course, but if you want to avoid https://warthogs.atlassian.net/browse/DPE-4855 you can add `raise_on_error=False` in this specific wait.

tests/integration/ha_tests/test_upgrade_to_primary_label.py
```diff
@@ -0,0 +1,135 @@
...
+        ops_test.model.deploy(
+            APPLICATION_NAME,
+            num_units=1,
+            channel="latest/edge",
+        ),
+    )
+    logger.info("Wait for applications to become active")
+    async with ops_test.fast_forward():
+        await ops_test.model.wait_for_idle(
+            apps=[DATABASE_APP_NAME, APPLICATION_NAME], status="active"
```
lucasgameiroborges, line 53: Same here

src/charm.py
```diff
@@ -937,12 +949,13 @@ def is_blocked(self) -> bool:
...
-            self._create_services()
-        except ApiError:
-            logger.exception("failed to create k8s services")
-            self.unit.status = BlockedStatus("failed to create k8s services")
-            return
+        if self.upgrade.idle:
+            try:
+                self._create_services()
+            except ApiError:
+                logger.exception("failed to create k8s services")
```
lucasgameiroborges, line 956: More of a question here, sorry :sweat_smile: but will `log.exception` (without the error as argument) display the error info for debugging purposes (eg. message, stack, etc)? I imagine so, just want to confirm so this doesn´t become a mistery later on!

lucasgameiroborges (approved):
LGTM! :rocket:

tests/integration/ha_tests/test_upgrade.py
dragomirp, line 1: The writes_increasing checks seem to be failing consistently on juju 3.

lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -129,12 +128,10 @@ def _connect_to_database(
...
-        for attempt in Retrying(stop=stop_after_attempt(10), wait=wait_fixed(3), reraise=True):
-            with attempt:
-                connection = psycopg2.connect(
-                    f"dbname='{database if database else self.database}' user='{self.user}' host='{host}'"
-                    f"password='{self.password}' connect_timeout=1"
-                )
+        connection = psycopg2.connect(
+            f"dbname='{database if database else self.database}' user='{self.user}' host='{host}'"
+            f"password='{self.password}' connect_timeout=1"
+        )
```
marceloneppel, line 134: Reverted due to failures in the async replication test (timeouts).

Issue #536, opened 2024-06-28
Charm conflicts with extension enabled via SQL


TeodorPt:
## Steps to reproduce
1. Deploy a postgresql-k8s charm.
2. Run `CREATE EXTENSION IF NOT EXISTS pg_trgm WITH SCHEMA pg_catalog;` in the resulting database, directly.
3. Create an index using the extension, e.g. `CREATE INDEX example_idx ON example USING gin(name gin_trgm_ops);`.

## Expected behavior
(?)

## Actual behavior
The extension is successfully created in the instance. Later, on various juju events, the charm actively tries to remove `pg_trgm` because it is not enabled via the `plugin_pg_trgm_enable` config. It however fails to do so because there's an index depending on it. This is an excerpt from the Postgres logs:
```
user=operator,db=<redacted>,app=[unknown],client=<redacted>,line=7 LOG:  statement: DROP EXTENSION IF EXISTS pg_trgm;
user=operator,db=<redacted>,app=[unknown],client=<redacted>,line=8 ERROR:  cannot drop extension pg_trgm because other objects depend on it
user=operator,db=<redacted>,app=[unknown],client=<redacted>,line=9 DETAIL:  index <redacted> depends on operator class gin_trgm_ops for access method gin
user=operator,db=<redacted>,app=[unknown],client=<redacted>,line=10 HINT:  Use DROP ... CASCADE to drop the dependent objects too.
user=operator,db=<redacted>,app=[unknown],client=<redacted>,line=11 STATEMENT:  DROP EXTENSION IF EXISTS pg_trgm;
```

## Versions
Juju CLI: 3.5.1-ubuntu-amd64
Juju agent: 3.4.2
Charm revision: 264, version 14.11

## Additional Context
I am not sure if this is actually a bug or just a peculiar behavior that we have observed. Nevertheless, reporting here in case there's anything that can be done to improve the experience.

marceloneppel:
Hi, @TeodorPt! Thanks for the report.

We'll be working to fix this issue in the next two weeks.

lucasgameiroborges:
Hi there @TeodorPt, this issue was addressed in #567, where we now have a blocked status in case where the charm tries to disable a plugin but it can't. In such a scenario, the user can proceed in 2 ways:

- Keep the blocking objects and enable the plugin via juju config,
- Delete the blocking objects and wait for `update-status` event for the plugin to get disabled,

[...]

PR #567, opened 2024-07-17
[DPE-4801] Block charm if plugin disable fails due to dependent objects

lucasgameiroborges:
## Issue

See https://github.com/canonical/postgresql-k8s-operator/issues/536

## Solution

If the charm detects objects that depend on plugins not enabled via config (most likely because the plugin was externally enabled) It will go into blocked state and warn the user. 

The user can then decide to either: Delete the objects and proceed with the plugin disabled (without updating the config) or update the juju config option of the plugin as enabled.

Local testing works, Will add unit/integration tests shortly.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -99,6 +99,7 @@ from utils import any_cpu_to_cores, any_memory_to_bytes, new_password
 logger = logging.getLogger(__name__)
 
 EXTENSIONS_DEPENDENCY_MESSAGE = "Unsatisfied plugin dependencies. Please check the logs"
+EXTENSION_OBJECT_MESSAGE = "Cannot disable plugins: Existing objects depend on it. See logs"
 
 # http{x,core} clutter the logs with debug messages
 logging.getLogger("httpcore").setLevel(logging.ERROR)
@@ -602,7 +603,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
                 )
                 return
 
-    def enable_disable_extensions(self, database: str = None) -> None:
+    def enable_disable_extensions(self, database: str = None) -> None:  # noqa: C901
         """Enable/disable PostgreSQL extensions set through config options.
 
         Args:
@@ -637,6 +638,13 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             self.unit.status = WaitingStatus("Updating extensions")
         try:
             self.postgresql.enable_disable_extensions(extensions, database)
+        except psycopg2.errors.DependentObjectsStillExist as e:
+            logger.error(
+                "Failed to disable plugin: %s Was the plugin enabled manually? If so, update charm config with `juju config postgresql-k8s plugin_<plugin name>_enable=True` ",
+                str(e),
+            )
+            self.unit.status = BlockedStatus(EXTENSION_OBJECT_MESSAGE)
+            return
         except PostgreSQLEnableDisableExtensionError as e:
             logger.exception("failed to change plugins: %s", str(e))
         if not isinstance(original_status, UnknownStatus):
--- a/src/charm.py
+++ b/src/charm.py
@@ -647,6 +647,9 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             return
         except PostgreSQLEnableDisableExtensionError as e:
             logger.exception("failed to change plugins: %s", str(e))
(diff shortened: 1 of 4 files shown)


marceloneppel (commented):
Thanks, Lucas! Nice UX!

Testing manually, I saw that the charm is not clearing the blocked status after the plugin is enabled.
```sql
CREATE EXTENSION IF NOT EXISTS pg_trgm WITH SCHEMA pg_catalog;
CREATE TABLE example(name VARCHAR);
CREATE INDEX example_idx ON example USING gin(name gin_trgm_ops);
```

```sh
juju config postgresql-k8s request_date_style="ISO, DMY"

# Wait for the charm do become blocked with the following message: Cannot disable plugins: Existing objects depend on it. See logs

juju config postgresql-k8s plugin_pg_trgm_enable=True

# Wait for the charm to become idle again. It's still blocked with the same message.
```

Also, will the charm become active again if the user deletes the index that uses the extension? I think we may need to have at least something that detects that.

lucasgameiroborges:
Thanks for the feedback @marceloneppel, I addressed the issues in the last commit, LMK if you still encounter them.

> Also, will the charm become active again if the user deletes the index that uses the extension? I think we may need to have at least something that detects that.

Just added a check inside `update_status`  that retries disabling the plugin. This was the fastest way to resolve the block in case the user decided to remove the dependent objects  themselves. Though I personally think waiting the ~5 minutes for the update_status call to resolve the block is a bit too much, but I don´t see any other way. Open to suggestions if you see a quicker way!

PS: Will add the missing tests once the behavior gets approved.

marceloneppel (approved):
LGTM!

The `update-status-hook` may take more than 5 minutes to fire in some environments because it's configurable.

As we don't have an observer like on the VM charm or any other mechanism right now, I think it's fine to keep the logic to set the active status back in the update status for now.

src/charm.py
```diff
@@ -602,7 +603,7 @@ def _on_config_changed(self, event) -> None:
                 )
                 return
 
-    def enable_disable_extensions(self, database: str = None) -> None:
+    def enable_disable_extensions(self, database: str = None) -> None:  # noqa: C901
```
dragomirp, line 606: I know cyclomatic complexity is annoying, but we should either bump the limit or fit within it, not disable it on individual functions.

src/charm.py
```diff
@@ -1239,6 +1255,9 @@ def _on_update_status(self, _) -> None:
             return
 
         if self._has_blocked_status or self._has_waiting_status:
+            # If charm was failing to disable plugin, try again (user may have removed the objects)
+            if self.unit.status.message == EXTENSION_OBJECT_MESSAGE:
+                self.enable_disable_extensions()
```
taurus-forever, line 1260: Do we want to call enable_disable_extensions() on every update-status?
This might slow-down PG with a lot of DBs.... which could be production.

IMHO, we can (should?) go into `error` state here and user have to call `juju resolve` after disabling of manually enabled extension.

taurus-forever (approved):
LGTM, please add tests. I would go different approach, but up 2 Marcelo here.

P.S. Do we have understanding how expensive `enable_disable_extensions()`?

src/charm.py
lucasgameiroborges: We only call it if the charm is in blocked status due to existing dependent objects. In normal workflow this shouldn't be an issue AFAICT, since 99.9% of the time the charm won't go inside the if statement. 

The problem I see with entering an `error` state (besides looking scary for the unsuspecting user) is that the charm doesn´t get the chance to auto-resolve: The most usual use-case (I think) is one where the user manually enables/uses an extension via SQL, and forgets to update juju config. They will then see the blocked status and logs and will go "oops lemme update this config my_plugin_enable=True" which will trigger the `config_changed` event and resolve itself.

This check inside juju status is for the (more unlikely) case where the user prefers to delete the objects and disable the extension. Here, if the charm sees that it was blocked due to dependent objects, it will retry removing the extension (in case the user deleted the offending objects).

That's why I would vouch for blocking the charm instead of going into error. @marceloneppel what do you think?

marceloneppel: After hearing many times that we should avoid the error state, I prefer to use a blocked status here (it explicitly calls for action from the user instead of giving the impression that the charm is failing if it gets into an error state). As Lucas commented, it's scary to see the error status in a charm.

dragomirp (approved):

marceloneppel (approved):

Issue #540, opened 2024-07-01
The postgresql peeble service went down and it wasn't restarted automatically


andrejvelichkovski:
Please see the Matrix thread here in case the full message context helps in understanding this issue ([link](https://matrix.to/#/!BukWfnyOTgQSKAxdtT:ubuntu.com/$N4MIkg3Te2YTF5RrPjZynvp2S5R5RZp7tKEkWek2xsU?via=ubuntu.com&via=matrix.org&via=laquadrature.net))

## Steps to reproduce

1. Deploy/run a `postgresql-k8s` charm
2. Kill the pebble `postgresql` service in the k8s pod.
3. The service doesn't restart itself automatically after it is killed unless manually triggered

## Expected behavior

We would expect that the service restarts itself automatically

## Actual behavior

The postgresql pebble service is inactive and doesn't restart itself.

## Versions

Operating system: PS5 environment

Juju CLI: 2.9.49-ubuntu-amd64

Juju agent: 14.11

Charm revision: 247

microk8s: 

## Log output

Juju debug log: 

```
2024-06-29T10:01:59.660Z [pebble] GET /v1/services?names=postgresql 207.373µs 200
2024-06-29T10:07:57.673Z [pebble] GET /v1/services?names=postgresql 279.914µs 200
2024-06-29T10:12:55.959Z [pebble] GET /v1/services?names=postgresql 379.654µs 200
2024-06-29T10:17:53.183Z [pebble] GET /v1/services?names=postgresql 318.372µs 200
2024-06-29T10:23:50.628Z [pebble] GET /v1/services?names=postgresql 291.239µs 200
2024-06-29T10:29:19.645Z [pebble] GET /v1/services?names=postgresql 176.82µs 200
2024-06-29T10:34:28.689Z [pebble] GET /v1/services?names=postgresql 328.981µs 200
2024-06-29T10:38:53.221Z [pebble] GET /v1/services?names=postgresql 185.35µs 200
2024-06-29T10:44:05.568Z [pebble] GET /v1/services?names=postgresql 314.562µs 200
2024-06-29T10:48:27.660Z [pebble] GET /v1/services?names=postgresql 270.051µs 200
2024-06-29T10:53:48.630Z [pebble] GET /v1/services?names=postgresql 226.33µs 200
2024-06-29T10:58:39.514Z [pebble] GET /v1/services?names=postgresql 223.46µs 200
[...]

PR #581, opened 2024-07-22
[DPE-4806] Restart pebble service if it's down

lucasgameiroborges:
## Issue

https://github.com/canonical/postgresql-k8s-operator/issues/540

## Solution

 - Check pebble status inside `update_status` hook. Restart if not enabled.
 - Update unit tests and integrations tests.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1280,6 +1280,17 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             logger.debug("on_update_status early exit: Service has not been added nor started yet")
             return
 
+        if (
+            "restoring-backup" not in self.app_peer_data
+            and services[0].current != ServiceStatus.ACTIVE
+        ):
+            logger.warning(
+                "on_update_status early exit: pebble service inactive, restarting container"
+            )
+            self.unit.status = MaintenanceStatus("Database service inactive, restarting")
+            container.restart(self._postgresql_service)
+            return
+
         if "restoring-backup" in self.app_peer_data and not self._was_restore_successful(
             services[0]
         ):
--- a/src/charm.py
+++ b/src/charm.py
@@ -1282,14 +1282,17 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
 
         if (
             "restoring-backup" not in self.app_peer_data
+            and "stopped" not in self.unit_peer_data
             and services[0].current != ServiceStatus.ACTIVE
         ):
             logger.warning(
                 "%s pebble service inactive, restarting service" % self._postgresql_service
             )
-            self.unit.status = MaintenanceStatus("Database service inactive, restarting")
             container.restart(self._postgresql_service)
-            return
+            # If service doesn't recover fast, exit and wait for next hook run to re-check
+            if not self._patroni.member_started:
+                self.unit.status = MaintenanceStatus("Database service inactive, restarting")
+                return
 
         if "restoring-backup" in self.app_peer_data and not self._was_restore_successful(
             services[0]
--- a/src/charm.py
+++ b/src/charm.py
@@ -1285,7 +1285,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             and services[0].current != ServiceStatus.ACTIVE
         ):
             logger.warning(
-                "on_update_status early exit: pebble service inactive, restarting container"
+                "%s pebble service inactive, restarting service" % self._postgresql_service
             )
             self.unit.status = MaintenanceStatus("Database service inactive, restarting")
             container.restart(self._postgresql_service)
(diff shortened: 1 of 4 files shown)


lucasgameiroborges:
Since we were already stopping and starting the pebble service manually in some integration tests, I removed the manual start step in those tests to allow for the auto-restart to take place. With this, I didn't see the need for a brand new integration test for this behavior. If  you still think we should have it, LMK!

lucasgameiroborges:
Whaaat, full green CI on first run!! So nice to see it 🥹

src/charm.py
```diff
@@ -1280,6 +1280,17 @@ def _on_update_status(self, _) -> None:
             logger.debug("on_update_status early exit: Service has not been added nor started yet")
             return
 
+        if (
+            "restoring-backup" not in self.app_peer_data
+            and services[0].current != ServiceStatus.ACTIVE
+        ):
+            logger.warning(
+                "on_update_status early exit: pebble service inactive, restarting container"
```
dragomirp, line 1288:
```suggestion
                "%s pebble service inactive, restarting service" % self._postgresql_service
```
We are restarting the service not the container. Since it's not a debug message we don't really need the early exit prefix.

src/charm.py
```diff
@@ -1280,6 +1280,17 @@ def _on_update_status(self, _) -> None:
...
+            )
+            self.unit.status = MaintenanceStatus("Database service inactive, restarting")
```
dragomirp, line 1290: Should we really change the status here? Assuming the `container.restart()` works out we should be back to active on the next line, while the new status will remain until another hook clears it up.

lucasgameiroborges: Supposing we don't use `return` inside this if statement, I guess the container restart would work and resolve the situation in time before the `set_active_status` by the end of the hook, the problem is: What if it doesn't resolve it in time? what if the restart doesn't fix the underlying problem? I thought about using this status just to let the user know what's happening, and not keep the normal active status in such worse cases.

One thing we could do is just removing the `return` inside the if, though I'm not sure it would make sense to proceed with the hook execution right after a container restart.

pinging @marceloneppel too

src/charm.py
```diff
@@ -1280,6 +1280,17 @@ def _on_update_status(self, _) -> None:
             logger.debug("on_update_status early exit: Service has not been added nor started yet")
             return
 
+        if (
+            "restoring-backup" not in self.app_peer_data
```
marceloneppel, line 1284: I believe async replication may also need to skip this logic.

https://github.com/canonical/postgresql-k8s-operator/blob/ea071d617bde602e0b168496b29e3758e3a47a63/src/relations/async_replication.py#L704

One way to check it is by checking whether the `stopped` flag is set in the relation data. It's being set after the container is stopped.

https://github.com/canonical/postgresql-k8s-operator/blob/ea071d617bde602e0b168496b29e3758e3a47a63/src/relations/async_replication.py#L718

tests/integration/ha_tests/test_self_healing.py
```diff
@@ -298,17 +298,9 @@ async def test_forceful_restart_without_data_and_transaction_logs(
...
-        logger.info(f"starting database on {primary_name}")
-        for attempt in Retrying(stop=stop_after_delay(30), wait=wait_fixed(3), reraise=True):
-            with attempt:
-                await run_command_on_unit(
-                    ops_test, primary_name, "/charm/bin/pebble start postgresql"
-                )
-
-                # Verify that the database service got restarted and is ready in the old primary.
-                logger.info(f"waiting for the database service to restart on {primary_name}")
-                assert await is_postgresql_ready(ops_test, primary_name)
```
marceloneppel, line 311: For this specific test, we should keep the logic that starts the Pebble service only after the WAL rotation is performed. Can you check if removing `async with ops_test.fast_forward():` or, even better, increasing the `update-status-hook-interval` can still make the test pass? My idea is to avoid the update status hook execution for this test specifically.

src/charm.py
marceloneppel: For this specific situation, @lucasgameiroborges, you may create a `tenacity.Retrying` block to check if the service started back up again (through the result from `self._patroni.member_started`). If it's okay, you can call `self._set_active_status()`. Otherwise, return and wait for the next hook that fixes the status.

src/charm.py
lucasgameiroborges: Honestly surprised this didn't cause the async_replication test to fail. In any case, nice catch, I'll add this check

src/charm.py
lucasgameiroborges: Implemented what you suggested @marceloneppel with the slight difference that, instead of explicitly calling `self._set_active_status()` I just let the hook finish and set the status in the final line, like normal. LMK if it makes sense.

marceloneppel (approved):

src/charm.py
marceloneppel: That's fine for me. Thanks!

dragomirp (approved):

taurus-forever (approved):
Tnx!

PR #595, opened 2024-07-29
[MISC] Switch test app interface

dragomirp:
Switch test-app interface from `first-database` to `database`


taurus-forever (approved):
Nice!

PR #602, opened 2024-07-31
[MISC] Adds missing `await` to `invalid_extra_user_roles` integration test + fix check loop

lucasgameiroborges:
## Issue

New relations integration test had a typo in `test_invalid_extra_user_roles` where two `block_until` calls were not being properly awaited. Once those were added, the test started failing due to blocked status resolving when it shouldn't have.

## Solution

- Add the necessary `await`s
- Use `continue` instead of `break` inside check loop to resolve the blocked status for invalid extra user roles.

This was previously done on VM inside https://github.com/canonical/postgresql-operator/pull/546

### DIFF
--- a/src/relations/postgresql_provider.py
+++ b/src/relations/postgresql_provider.py
@@ -252,7 +252,7 @@ class PostgreSQLProvider(Object):
             for data in relation.data.values():
                 extra_user_roles = data.get("extra-user-roles")
                 if extra_user_roles is None:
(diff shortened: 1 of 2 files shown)


taurus-forever (approved):
Nice catch!

dragomirp (approved):

marceloneppel (approved):
Thanks!

PR #623, opened 2024-08-12
[MISC] Bump coverage

dragomirp:
* Increase the coverage so that codecov stops complaining
* Update libs (except `kubernetes_service_patch`)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,6 +86,9 @@ branch = true
 
 [tool.coverage.report]
 show_missing = true
+exclude_lines = [
+    "logger\\.debug"
+]
 
 [tool.pytest.ini_options]
 minversion = "6.0"
--- a/src/patroni.py
+++ b/src/patroni.py
@@ -12,6 +12,7 @@ from typing import Any, Dict, List, Optional
 import requests
 import yaml
 from jinja2 import Template
+from ops.pebble import Error
 from tenacity import (
     AttemptManager,
     RetryError,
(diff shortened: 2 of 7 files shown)


pyproject.toml
```diff
@@ -86,6 +86,9 @@ branch = true
 
 [tool.coverage.report]
 show_missing = true
+exclude_lines = [
+    "logger\\.debug"
+]
```
dragomirp, line 91: Excluded on VM: https://github.com/canonical/postgresql-operator/blob/main/pyproject.toml#L84-L86

src/patroni.py
```diff
@@ -476,16 +477,16 @@ def last_postgresql_logs(self) -> str:
...
-        log_files.sort(key=lambda f: f.path, reverse=True)
         try:
+            log_files = container.list_files(POSTGRESQL_LOGS_PATH, pattern=POSTGRESQL_LOGS_PATTERN)
+            if len(log_files) == 0:
+                return ""
+            log_files.sort(key=lambda f: f.path, reverse=True)
             with container.pull(log_files[0].path) as last_log_file:
                 return last_log_file.read()
-        except OSError as e:
+        except Error:
```
dragomirp, line 487: I don't think accessing the container will raise an `OSError` and list files will error if the directory doesn't exist.

src/relations/postgresql_provider.py
```diff
@@ -78,9 +78,6 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
             event.defer()
             return
 
-        if not self.charm.unit.is_leader():
-            return
-
```
dragomirp, line 83: Dead code, database_requested should only be emitted on the leader: https://github.com/canonical/data-platform-libs/blob/main/lib/charms/data_platform_libs/v0/data_interfaces.py#L2878-L2879

tests/unit/test_patroni.py
dragomirp, line 1: Mostly copypaste from the VM charm.

dragomirp, line 1: Mostly copypaste from the VM charm.

taurus-forever (approved):
Tnx!

marceloneppel (approved):
Thanks!

PR #624, opened 2024-08-12
[MISC] service patch lib update

dragomirp:
Update service patch charm lib to latest version.


taurus-forever (approved):
Wow! This is a really nice catch!

marceloneppel (approved):

PR #626, opened 2024-08-13
[DPE-3562] Don't block on missing Postgresql version

dragomirp:

### DIFF
--- a/src/relations/db.py
+++ b/src/relations/db.py
@@ -152,12 +152,9 @@ class DbProvides(Object):
             self.charm.unit.status = BlockedStatus(ROLES_BLOCKING_MESSAGE)
             return False
 
-        database = relation.data.get(relation.app, {}).get("database")
-        if not database:
+        if not (database := relation.data.get(relation.app, {}).get("database")):
             for unit in relation.units:
-                unit_database = relation.data.get(unit, {}).get("database")
-                if unit_database:
-                    database = unit_database
+                if database := relation.data.get(unit, {}).get("database"):
                     break
 
         if not database:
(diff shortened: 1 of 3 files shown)


src/relations/db.py
```diff
@@ -207,31 +204,40 @@ def set_up_relation(self, relation: Relation) -> bool:
...
+                "standbys": standbys,
+                "user": user,
+                "password": password,
+                "database": database,
+                "extensions": ",".join(required_extensions),
+            }
+            if postgresql_version:
+                updates["version"] = postgresql_version
+            application_relation_databag.update(updates)
+            unit_relation_databag.update(updates)
```
dragomirp, line 237: So that that mccabe doesn't complain.

src/relations/db.py
```diff
@@ -207,31 +204,40 @@ def set_up_relation(self, relation: Relation) -> bool:
...
         except (
             PostgreSQLCreateDatabaseError,
             PostgreSQLCreateUserError,
-            PostgreSQLGetPostgreSQLVersionError,
         ):
             self.charm.unit.status = BlockedStatus(
```
dragomirp, line 242: We should most probably defer here. Should I change this as well?

src/relations/db.py
```diff
@@ -207,31 +204,40 @@ def set_up_relation(self, relation: Relation) -> bool:
...
 
+            postgresql_version = None
+            try:
+                postgresql_version = self.charm.postgresql.get_postgresql_version()
+            except PostgreSQLGetPostgreSQLVersionError:
+                logger.exception(
+                    "Failed to retrieve the PostgreSQL version to initialise/update %s relation"
+                    % self.relation_name
+                )
+
```
dragomirp, line 215: K8S charm will not update the endpoints as aggressively as VM. This will be reran only on config change. Should I try add something to update status?

taurus-forever (approved):
Up 2 Marcelo...

src/relations/db.py
marceloneppel: If it's possible to avoid calls to `sleep` or similar things on the notice event received by the charm (if we need to notify again from inside that event handler because the database version couldn't still be retrieved), I'm fine with using Pebble notices for that. Otherwise, I'd keep it on update status for now.

src/relations/db.py
marceloneppel: I think a `defer` would be good here. We just need to ensure that we can change the status to `Active` later when the database/user creation succeeds (which is what we are already missing today).

marceloneppel (approved):

src/relations/db.py
```diff
@@ -96,7 +96,13 @@ def _on_relation_changed(self, event: RelationChangedEvent) -> None:
...
         logger.warning(f"DEPRECATION WARNING - `{self.relation_name}` is a legacy interface")
 
-        self.set_up_relation(event.relation)
+        if (
+            not self.set_up_relation(event.relation)
+            and self.charm.unit.status.message
+            == f"Failed to initialize {self.relation_name} relation"
+        ):
+            event.defer()
+            return
```
dragomirp, line 105: Charm should be unblocked by `_update_unit_status` if set_up_relation succeeds.

marceloneppel (approved):

taurus-forever (approved):

Issue #638, opened 2024-08-19
`JujuVersion` should be imported from `ops.jujuversion` instead of `ops.model`


IronCore864:
[Here](https://github.com/canonical/postgresql-k8s-operator/blob/main/src/charm.py#L63) `JujuVersion` is imported from `ops.model` (although it does work, since `ops.model` also imports `ops.jujuversion` for now), it should import `JujuVersion` from `ops.jujuversion`; see [here](https://github.com/canonical/operator/blob/main/ops/__init__.py#L123).

PR #640, opened 2024-08-19
bugfix: import JujuVersion from ops.jujuversion instead of ops.model

IronCore864:
## Issue

https://github.com/canonical/postgresql-k8s-operator/issues/638

## Solution

Import `JujuVersion` from `ops.jujuversion` instead of `ops.model`.

Current, importing `JujuVersion` from `ops.model` works but only because `ops.model` happens to import `JujuVersion from `ops.jujuversion`. This will be changed soon, so we should import `JujuVersion` from `ops.jujuversion`.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -55,12 +55,12 @@ from ops.charm import (
     RelationDepartedEvent,
     WorkloadEvent,
 )
+from ops.jujuversion import JujuVersion
 from ops.main import main
 from ops.model import (
     ActiveStatus,
     BlockedStatus,
     Container,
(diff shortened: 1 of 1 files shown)


dragomirp (approved):

taurus-forever (approved):
Thank you for the contribution!

IronCore864:
Sorry to bother you again, but after discussion, we will do some refactor about `jujuversion.py` as well, so it will be `from ops import JujuVersion` instead of `from ops.jujuversion import JujuVersion`. I will do another PR to fix this.

PR #661, opened 2024-08-28
[DPE-5275] patroni passwords

dragomirp:
Add password to Patroni's REST API.

Partial port of https://github.com/canonical/postgresql-operator/pull/596


metadata.yaml
```diff
@@ -28,7 +28,7 @@ resources:
   postgresql-image:
     type: oci-image
     description: OCI image for PostgreSQL
-    upstream-source: ghcr.io/canonical/charmed-postgresql@sha256:7ef86a352c94e2a664f621a1cc683d7a983fd86e923d98c32b863f717cb1c173 # renovate: oci-image tag: 14.12-22.04_edge
+    upstream-source: ghcr.io/canonical/charmed-postgresql@sha256:2c066876e80d60058d79835c8b5d18090963b3a0f84261385afa1fc652477605 # renovate: oci-image tag: 14.12-22.04_edge
```
dragomirp, line 31: This should be the last 14.12 image

src/upgrade.py
```diff
@@ -143,12 +144,17 @@ def _on_postgresql_pebble_ready(self, event: WorkloadEvent) -> None:
...
         """Update the Patroni nosync tag in the unit if needed."""
         if not self.peer_relation or not self.charm._patroni.member_started:
             return
 
         self.charm.update_config()
+        container = self.charm.unit.get_container("postgresql")
+        if not container.can_connect():
+            event.defer()
+            return
+        container.send_signal(SIGHUP, "postgresql")
```
dragomirp, line 157: Secret will be generated during the Juju leader upgrade, so we should reload the config for units that upgraded earlier to pick it up.

taurus-forever: It sounds hackish, but no better proposal right now.

taurus-forever (approved):
LGTM.

src/upgrade.py
dragomirp: This should be equivalent to calling the `/reload` endpoint and it shouldn't restart the DB again. We can't use the REST API or `patronictl` since creds will be in flux during the upgrade.

marceloneppel (approved):

PR #663, opened 2024-08-29
[MISC] Temporarily disable log forwarding & fix for race in Patroni REST password setup

dragomirp:
* Temporarily disable log forwarding to be able to release 14.12 and Patroni auth without logging degradation.
* Only set Patroni's REST auth if the secret is available


taurus-forever (approved):
Thank you for the small revert approach!

marceloneppel (approved):

tests/integration/helpers.py
```diff
@@ -744,15 +744,17 @@ async def switchover(
...
+        with attempt:
+            response = requests.post(
+                f"http://{primary_ip}:8008/switchover",
+                json={
+                    "leader": current_primary.replace("/", "-"),
+                    "candidate": candidate.replace("/", "-") if candidate else None,
+                },
+                auth=requests.auth.HTTPBasicAuth("patroni", password),
+            )
+            assert response.status_code == 200, f"Switchover status code is {response.status_code}"
```
dragomirp, line 757: Retrying the switchover, in case the REST pass is still being set up.

src/upgrade.py
```diff
@@ -150,11 +149,6 @@ def _on_upgrade_changed(self, event) -> None:
             return
 
         self.charm.update_config()
-        container = self.charm.unit.get_container("postgresql")
-        if not container.can_connect():
-            event.defer()
-            return
-        container.send_signal(SIGHUP, "postgresql")
```
dragomirp, line 157: The reload in update_config should work now.

PR #664, opened 2024-08-29
[DPE-5244] Split PITR backup test in AWS and GCP

lucasgameiroborges:
## Issue

PITR backup tests should run in parallel on both GCP and AWS clouds.

## Solution

- Refactor (similar to non-PITR backup tests) to a helper function.
- 2 tests in different groups.


tests/integration/helpers.py
```diff
@@ -1029,3 +1030,197 @@ async def backup_operations(
...
+async def pitr_backup_operations(
+    ops_test: OpsTest,
+    s3_integrator_app_name: str,
+    tls_certificates_app_name: str,
+    tls_config,
+    tls_channel,
+    credentials,
+    cloud,
+    config,
+) -> None:
```
dragomirp, line 1044: IMHO this should stay in the PITR test as a hanging functions, since it's not really reusable by anything else.

dragomirp (approved):

tests/integration/helpers.py
marceloneppel: I agree. We can move this now to there or later—no need to block the PR.

marceloneppel (approved):
LGTM!

tests/integration/helpers.py
lucasgameiroborges: Should be better now, lighter diff too :)

taurus-forever (approved):

PR #613, opened 2024-08-06
[DPE-4239] Warn user when deploying charm with wrong architecture

lucasgameiroborges:
## Issue

When deploying a revision that doesn't correspond to the underlying architecture (amd/arm) the charm deployment bugs out with import errors instead of properly warning the user.

## Solution

Not very pretty - Try-catching the first platform specific library (psycopg2 in our case) and creating a fake charm blocked with the error message.

I personally don't like it, but that's the only working solution we have so far, since the charm code will run the import statements before everything else. Let me know if you think it's something worth adding.

## Tests

Unit and Integration tests added

PR #665, opened 2024-08-29
[DPE-4239] Use manifest file to check for charm architecture

lucasgameiroborges:
## Issue
Fixes a bug reported on https://github.com/canonical/postgresql-k8s-operator/pull/613#issuecomment-2294369041 where the import error would get re-raised because the architecture check was not working properly (returning `False` due to not finding the `.juju-charm` file). 

## Solution

Use `manifest.yaml` instead of `.juju-charm`, it is a more reliable check for some reason.

## Test

CI pass (always have been, apparently jujus 2.9 and 3.4 never had the problem?) + local tests on AMD and juju 3.5 work + @marceloneppel tested on ARM + juju 3.5 and also worked. Should be fine now!

### DIFF
--- a/src/arch_utils.py
+++ b/src/arch_utils.py
@@ -26,7 +26,7 @@ class WrongArchitectureWarningCharm(CharmBase):
 
 def is_wrong_architecture() -> bool:
     """Checks if charm was deployed on wrong architecture."""
-    juju_charm_file = f"/var/lib/juju/agents/{os.environ.get('HOSTNAME')}/charm/.juju-charm"
+    juju_charm_file = f"/var/lib/juju/agents/unit-{os.environ.get('HOSTNAME')}/charm/.juju-charm"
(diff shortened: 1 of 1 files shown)


taurus-forever (approved):

dragomirp:
Is `.jujucharm` no longer created on 3.5?

dragomirp (approved):

lucasgameiroborges:
> Is `.jujucharm` no longer created on 3.5?

It was there when I ssh'd into the pod, but the arch check on install was failing due to not finding it. So, maybe it's creation got delayed? In any case, I think its better to rely on the manifest file, as it's purpose is explicitly to contain this sort of info: https://juju.is/docs/sdk/manifest-yaml

marceloneppel (approved):
LGTM! Thanks!

PR #668, opened 2024-08-30
[TEST] Bump timeout

dragomirp:


dragomirp:
Looks like a race in the Patroni REST password setup

PR #670, opened 2024-08-30
[MISC] Conditional password

dragomirp:

### DIFF
--- a/tests/unit/test_upgrade.py
+++ b/tests/unit/test_upgrade.py
@@ -1,6 +1,5 @@
 # Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
-from signal import SIGHUP
 from unittest.mock import MagicMock, PropertyMock, call, patch
 
 import pytest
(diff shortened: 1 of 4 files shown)


tests/unit/test_upgrade.py
```diff
@@ -170,7 +168,6 @@ def test_on_upgrade_changed(harness):
         _member_started.return_value = True
         harness.charm.on.upgrade_relation_changed.emit(relation)
         _update_config.assert_called_once()
-        _send_signal.assert_called_once_with(SIGHUP, "postgresql")
```
taurus-forever, line 173: Why do you remove it?

taurus-forever (approved):
LGTM as Dragomir and test are happy.
IMHO, it worth to show this Marcelo, or at least @lucasgameiroborges as @marceloneppel is on vacation today.

tests/unit/test_upgrade.py
dragomirp: > Why do you remove it?

Because now the reload REST call in `update_config()` will work during the upgrade and we no longer do the additional SIGHUP.

dragomirp:
Superseded by https://github.com/canonical/postgresql-k8s-operator/pull/663

PR #671, opened 2024-08-31
[MISC] Re-enable log forwarding

dragomirp:
* Re-enable pebble log forwarding
* Update libs


taurus-forever (approved):
Thank you for the quick revert!

marceloneppel (approved):

Issue #669, opened 2024-08-30
Patroni Postgresql instance is down due to k8s api unavailable


cbartz:
## Steps to reproduce

1. Deploy postgresql-k8s with three units.
2. Integrate it with another k8s app. 

Example deployment: https://pastebin.canonical.com/p/YRyVgZpxC5/

## Expected behavior

Cluster does not fail on K8s API unavailability.

## Actual behavior

We received the alert

```
[[FIRING:1] PatroniPostgresqlDown (alertname="PatroniPostgresqlDown", instance="prod-repo-policy-compliance_e3901794-73c7-43f0-85ae-5179a6dc1b16_postgresql_postgresql/0", job="juju_prod-repo-policy-compliance_e3901794_postgresql_prometheus_scrape-0_60d7da0f2146e3c09ca949795f3ca709e4494a27fcf14b443a684b940f4d44fe", juju_charm="postgresql-k8s", juju_unit="postgresql/0", scope="patroni-postgresql", severity="critical")](http://alertmanager-0.alertmanager-endpoints.prod-cos-k8s-ps6-is-charms.svc.cluster.local:9093/#/alerts?receiver=mattermost-notifications)
Alert: Patroni Posrgresql Down (instance prod-repo-policy-compliance_e3901794-73c7-43f0-85ae-5179a6dc1b16_postgresql_postgresql/0) - critical

Description: Patroni Postgresql instance is down
  VALUE = 0
  LABELS = map[name:patroni_postgres_running instance:prod-repo-policy-compliance_e3901794-73c7-43f0-85ae-5179a6dc1b16_postgresql_postgresql/0 job:juju_prod-repo-policy-compliance_e3901794_postgresql_prometheus_scrape-0_60d7da0f2146e3c09ca949795f3ca709e4494a27fcf14b443a684b940f4d44fe juju_application:postgresql juju_charm:postgresql-k8s juju_model:prod-repo-policy-compliance juju_model_uuid:e3901794-73c7-43f0-85ae-5179a6dc1b16 juju_unit:postgresql/0 scope:patroni-postgresql]

Details:
   • alertname: PatroniPostgresqlDown
   • instance: prod-repo-policy-compliance_e3901794-73c7-43f0-85ae-5179a6dc1b16_postgresql_postgresql/0
[...]

Issue #616, opened 2024-08-07
 Postgresql goes down due to "terminating connection due to administrator command"


cbartz:
## Steps to reproduce

Deploy postgresql 14/edge on rev 239 with three units.

```
Model                        Controller                         Cloud/Region              Version  SLA          Timestamp
prod-repo-policy-compliance  juju-controller-35-production-ps6  k8s-prod-general/default  3.5.3    unsupported  08:07:59Z

App                       Version  Status  Scale  Charm                     Channel      Rev  Address        Exposed  Message
postgresql                14.12    active      3  postgresql-k8s            14/edge      239  10.87.6.161    no       Primary
repo-policy-compliance             active      1  repo-policy-compliance    latest/edge   17  10.87.26.0     no       

Unit                         Workload  Agent  Address          Ports  Message
postgresql/0                 active    idle   192.168.102.57          Primary
postgresql/1*                active    idle   192.168.100.212         
postgresql/2                 active    idle   192.168.103.49   
repo-policy-compliance/0*    active    idle   192.168.102.43          
```

## Expected behavior

Postgreql is not going down.

## Actual behavior

We received the Postgresql down alert on a production deployment with three units.

```
[[FIRING:2] (instance="prod-repo-policy-compliance_e3901794-73c7-43f0-85ae-5179a6dc1b16_postgresql_postgresql/1", job="juju_prod-repo-policy-compliance_e3901794_postgresql_prometheus_scrape-1_638a329e43b79c2385aeef8f96e0df8ed4864b1b4128b49c0b1215c36f0adff0", juju_charm="postgresql-k8s", juju_unit="postgresql/1", severity="critical")](http://alertmanager-0.alertmanager-endpoints.prod-cos-k8s-ps6-is-charms.svc.cluster.local:9093/#/alerts?receiver=mattermost-notifications)
Alert: Postgresql down (instance prod-repo-policy-compliance_e3901794-73c7-43f0-85ae-5179a6dc1b16_postgresql_postgresql/1) - critical

Description: Postgresql instance is down
  VALUE = 0
[...]

PR #677, opened 2024-09-03
[DPE-5311] Enable DCS failsafe mode

dragomirp:
## Issue
Patroni depends on the K8s API and resources to maintain the leader lock. Primary will step down during a temporary K8s API outage

## Solution
Enable DCS failsafe mode. If the K8s API (DCS) is not available, Patroni will try to maintain the current primary if it can connect to all cluster members

Documentation: https://patroni.readthedocs.io/en/latest/dcs_failsafe_mode.html#dcs-failsafe-mode

Related #669 and #616

### DIFF
--- a/templates/patroni.yml.j2
+++ b/templates/patroni.yml.j2
@@ -1,6 +1,7 @@
 bootstrap:
   dcs:
     synchronous_mode: true
+    failsafe_mode: true
     synchronous_node_count: {{ minority_count }}
     postgresql:
       use_pg_rewind: true


delgod (approved):

delgod (approved):
after careful check of implementation details - everything looks safe (even in upgrade case) and I believe we should enable this option.

taurus-forever (approved):
Mykola approved, Mohamed notified in MM. LGTM from me. Thank you!

taurus-forever:
@7annaba3l please LGTM!

7annaba3l:
LGTM. Would be good if we can add a test for this specif scenario, though.
Please also double check the version we currently use for Patroni.

taurus-forever:
re: `Please also double check the version we currently use for Patroni.`

[patroni=3.1.2-0ubuntu0.22.04.1~ppa2
](https://github.com/canonical/charmed-postgresql-snap/blob/14/edge/snap/snapcraft.yaml#L238)

taurus-forever:
re: `Please also double check the version we currently use for Patroni.`

We are using `3.1.2-0ubuntu0.22.04.1~ppa2`.

Failsafe introduced in 3.0.0.
There are fixes in 3.1.1 (included).
Improved in 4.0.0 (not available for us).

(c) https://github.com/patroni/patroni/blob/master/docs/releases.rst

Issue #634, opened 2024-08-15
Missing `cluster-initialised` flag causes cluster to become stuck


AmberCharitos:
## Steps to reproduce

This was seen in production, the below is what happened in production. This does not necessarily mean it will reproduce the same bug locally as there may have been other factors at play. 

1. Deploy 3 units of postgresql-k8s charm channel 14/edge revision 198.
2. Restart all pods the same time.

## Expected behavior
All units come back to an active status

## Actual behavior
All units enter either maintenance or waiting status. With the following, cluster not initialized log message.
```
Re-emitting deferred event <PebbleReadyEvent via PostgresqlOperatorCharm/on/postgresql_pebble_ready[6552765]>.
unit-postgresql-db-0: 08:13:00 DEBUG unit.postgresql-db/0.juju-log Deferring on_postgresql_pebble_ready: Not leader and cluster not initialized

```

## Versions

Operating system: Ubuntu 22.04.4 LTS

Juju CLI: 3.5.3-ubuntu-amd64

Juju agent: 3.1.8

Charm revision: `198` `14/edge`

kubectl:
Client Version: v1.30.3
Server Version: v1.26.15

## Log output

Juju debug log: 

[postgresql-0.log](https://github.com/user-attachments/files/16623037/postgresql-0.log)
[postgresql-1.log](https://github.com/user-attachments/files/16623039/postgresql-1.log)
[postgresql-2.log](https://github.com/user-attachments/files/16623043/postgresql-2.log)

## Additional context

[Matrix conversation for context] (https://matrix.to/#/!BukWfnyOTgQSKAxdtT:ubuntu.com/$jrGfGfW07m4i6t0dKcKAElmZmEznSvDlh1ruRTerFgI?via=ubuntu.com&via=matrix.org&via=laquadrature.net)

marceloneppel:
Thanks for the bug report, @AmberCharitos!

Just to not forget, the flag is removed at https://github.com/canonical/postgresql-k8s-operator/blob/e760a3df8ceceab49357ff57c620299d9d101fc8/src/charm.py#L841.

lucasgameiroborges:
Hi @AmberCharitos, thanks for the bug report!

Question: Have you observed this issue on newer charm revisions? I haven't been able to reproduce this issue locally, the charm seems to consistently recover in around 2 minutes after doing `kubectl delete pod pg-0 pg-1 pg-2`. Do you have specific steps?

[...]

PR #679, opened 2024-09-03
[DPE-5165] Add log for fix_leader_annotation method

marceloneppel:
## Issue
We don't know all the moments when the `leader` annotation is fixed (we know that one situation - expected - is after a scale down to 0 units and then a scale up to 1 unit), which causes the `cluster_initialised` flag to be removed from the peer relation data (this is causing issues in production like the one reported at https://github.com/canonical/postgresql-k8s-operator/issues/634).

## Solution
Add a log for the `fix_leader_annotation` method so we can better understand why the issue occurs.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -849,6 +849,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
                     Endpoints, name=self.cluster_name, namespace=self._namespace, obj=patch
                 )
                 self.app_peer_data.pop("cluster_initialised", None)
+                logger.info("Fixed missing leader annotation")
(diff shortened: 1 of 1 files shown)


lucasgameiroborges (approved):
LGTM, maybe include `self._unit` too for more info/context?

dragomirp (approved):

marceloneppel:
> LGTM, maybe include `self._unit` too for more info/context?

It already appears in the log before the logged messages. It will show something like:
```sh
unit-postgresql-k8s-0: 20:42:27 INFO unit.postgresql-k8s/0.juju-log Fixed missing leader annotation
```

Issue #433, opened 2024-03-26
Cluster lost files from data dir from the former primary and its replication slot


marceloneppel:
## Steps to reproduce
1. Deploy 3 PostgreSQL units.
2. Restart all the units through an outage.

## Expected behavior
All the units should have recovered from the outage.

## Actual behavior
Cluster lost files from data dir from the former primary and its replication slot.

## Versions
Charm revision: 198

## Log output
Patroni logs after the outage (in the former primary unit):
```
2024-03-22 14:26:44 UTC [196]: ERROR: Error when calling pg_controldata 
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/patroni/postgresql/__init__.py", line 999, in controldata
    data = subprocess.check_output([self.pgcommand('pg_controldata'), self._data_dir], env=env)
  File "/usr/lib/python3.10/subprocess.py", line 421, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.10/subprocess.py", line 526, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/usr/lib/postgresql/14/bin/pg_controldata', '/var/lib/postgresql/data/pgdata']' returned non-zero exit status 1.
2024-03-22 14:57:36 UTC [196]: ERROR: Error when calling pg_controldata
```

PostgreSQL logs after calling `curl {former-primary-ip}:8008/reinitialize` (in the former primary unit):
```
2024-03-22 15:36:04 UTC [645]: user=,db=,app=,client=,line=1 FATAL:  could not start WAL streaming: ERROR:  replication slot "postgresql_db_0" does not exist
2024-03-22 15:36:10 UTC [630]: user=,db=,app=,client=,line=7 LOG:  received immediate shutdown request
2024-03-22 15:36:10 UTC [630]: user=,db=,app=,client=,line=8 LOG:  database system is shut down
2024-03-22 15:36:24 UTC [754]: user=,db=,app=,client=,line=3 LOG:  starting PostgreSQL 14.10 (Ubuntu 14.10-0ubuntu0.22.04.1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0, 64-bit
2024-03-22 15:36:24 UTC [754]: user=,db=,app=,client=,line=4 LOG:  listening on IPv4 address "0.0.0.0", port 5432
[...]

PR #680, opened 2024-09-03
[DPE-3887] Avoid replication slot deletion

marceloneppel:
## Issue
After restarting the pods of a PostgreSQL K8s charm deployment (we can use `kubectl -n dev delete pod -l app.kubernetes.io/name=postgresql-k8s` for testing), if the `stop` hook fails for some reason, the `upgrade-charm` hook won't fire when the pod is rescheduled, and no Patroni labels will be added to the pod. This makes Patroni understand that the unit is not part of the cluster, so it deletes that unit replication slot from the primary.

In the past (before https://github.com/canonical/postgresql-k8s-operator/pull/448, which added a check for `container.can_connect()`), this could happen due to a re-emitted deferred pebble-ready event:
```sh
unit-postgresql-k8s-2: 17:36:20 ERROR unit.postgresql-k8s/2.juju-log Uncaught exception while in charm code:
  File "/var/lib/juju/agents/unit-postgresql-k8s-2/charm/./src/charm.py", line 1574, in <module>
    main(PostgresqlOperatorCharm, use_juju_for_storage=True)
  File "/var/lib/juju/agents/unit-postgresql-k8s-2/charm/venv/ops/main.py", line 434, in main
    framework.reemit()
  File "/var/lib/juju/agents/unit-postgresql-k8s-2/charm/venv/ops/framework.py", line 863, in reemit
    self._reemit()
  File "/var/lib/juju/agents/unit-postgresql-k8s-2/charm/venv/ops/framework.py", line 942, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-postgresql-k8s-2/charm/./src/charm.py", line 660, in _on_postgresql_pebble_ready
    self._create_pgdata(container)
  File "/var/lib/juju/agents/unit-postgresql-k8s-2/charm/./src/charm.py", line 646, in _create_pgdata
    if not container.exists(path):
  File "/var/lib/juju/agents/unit-postgresql-k8s-2/charm/venv/ops/model.py", line 2547, in exists
    self._pebble.list_files(str(path), itself=True)
  File "/var/lib/juju/agents/unit-postgresql-k8s-2/charm/venv/ops/pebble.py", line 2219, in list_files
    resp = self._request('GET', '/v1/files', query)
  File "/var/lib/juju/agents/unit-postgresql-k8s-2/charm/venv/ops/pebble.py", line 1655, in _request
    response = self._request_raw(method, path, query, headers, data)
  File "/var/lib/juju/agents/unit-postgresql-k8s-2/charm/venv/ops/pebble.py", line 1704, in _request_raw
    raise ConnectionError(
ops.pebble.ConnectionError: Could not connect to Pebble: socket not found at '/charm/containers/postgresql/pebble.socket' (container restarted?)
unit-postgresql-k8s-1: 17:36:20 INFO unit.postgresql-k8s/1.juju-log Updating Patroni config file
unit-postgresql-k8s-0: 17:36:20 INFO juju.worker.uniter.operation ran "stop" hook (via hook dispatching script: dispatch)
unit-postgresql-k8s-0: 17:36:20 INFO juju.worker.uniter unit "postgresql-k8s/0" shutting down: agent should be terminated
unit-postgresql-k8s-1: 17:36:20 INFO unit.postgresql-k8s/1.juju-log Updated health checks
unit-postgresql-k8s-2: 17:36:20 ERROR juju.worker.uniter.operation hook "stop" (via hook dispatching script: dispatch) failed: exit status 1
```

## Solution
Patch the pods in the `postgresql-pebble-ready` hook (only when the unit is already a member of the Patroni/PostgreSQL cluster), which is fired after the charm starts again. An integration test (forcing a failure in the `stop` hook) was added to avoid regressions.

Some safeguards still need to be added to the stop hook as a different situation may cause it to fail.

Fixes https://github.com/canonical/postgresql-k8s-operator/issues/433.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1027,25 +1027,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         return isinstance(self.unit.status, BlockedStatus)
 
     def _on_upgrade_charm(self, _) -> None:
-        # Recreate k8s resources and add labels required for replication
-        # when the pod loses them (like when it's deleted).
-        if self.upgrade.idle:
-            try:
-                self._create_services()
-            except ApiError:
-                logger.exception("failed to create k8s services")
-                self.unit.status = BlockedStatus("failed to create k8s services")
-                return
-
-        try:
-            self._patch_pod_labels(self.unit.name)
-        except ApiError as e:
-            logger.error("failed to patch pod")
-            self.unit.status = BlockedStatus(f"failed to patch pod with error {e}")
-            return
-
-        # Update the sync-standby endpoint in the async replication data.
-        self.async_replication.update_async_replication_data()
+        self._fix_pod()
 
     def _patch_pod_labels(self, member: str) -> None:
         """Add labels required for replication to the current pod.
@@ -1257,7 +1239,15 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         except RetryError as e:
             logger.error(f"failed to get primary with error {e}")
 
-    def _on_start(self):
+    def _on_start(self, _) -> None:
+        if self._endpoint not in self._endpoints:
+            return
+
+        self._fix_pod()
+
+    def _fix_pod(self) -> None:
+        # Recreate k8s resources and add labels required for replication
+        # when the pod loses them (like when it's deleted).
         if self.upgrade.idle:
             try:
                 self._create_services()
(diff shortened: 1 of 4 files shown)


dragomirp (approved):

lucasgameiroborges (approved):

taurus-forever (approved):
Wow! This is an excellent finding! Thank you!

PR #685, opened 2024-09-04
[DPE-5301] Add check for low storage space on pgdata volume

lucasgameiroborges:
## Issue

The charm gives no indication of low storage space on `pgdata` folder, which can cause unexpected failures

## Solution

Added check on `update_status` hook + integration test.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -941,6 +941,9 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         self._set_active_status()
 
     def _set_active_status(self):
+        # The charm should not override this status outside of the function checking disk space.
+        if self.unit.status.message == INSUFFICIENT_SIZE_WARNING:
+            return
         if "require-change-bucket-after-restore" in self.app_peer_data:
             if self.unit.is_leader():
                 self.app_peer_data.update({
@@ -1346,6 +1349,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         if free_size / total_size < 0.1:
             self.unit.status = BlockedStatus(INSUFFICIENT_SIZE_WARNING)
         elif self.unit.status.message == INSUFFICIENT_SIZE_WARNING:
+            self.unit.status = ActiveStatus()
             self._set_active_status()
 
     def _on_update_status(self, _) -> None:
--- a/src/charm.py
+++ b/src/charm.py
@@ -125,7 +125,7 @@ logger = logging.getLogger(__name__)
 
 EXTENSIONS_DEPENDENCY_MESSAGE = "Unsatisfied plugin dependencies. Please check the logs"
 EXTENSION_OBJECT_MESSAGE = "Cannot disable plugins: Existing objects depend on it. See logs"
-INSUFFICIENT_SIZE_WARNING = "<10%% free space on pgdata volume."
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -1327,12 +1329,33 @@ def _on_update_status_early_exit_checks(self, container) -> bool:
...
+        logger.debug(
+            "pgdata free disk space: %s out of %s, ratio of %s",
+            free_size,
+            total_size,
+            free_size / total_size,
+        )
+        if free_size / total_size < 0.1:
+            self.unit.status = BlockedStatus(INSUFFICIENT_SIZE_WARNING)
+        elif self.unit.status.message == INSUFFICIENT_SIZE_WARNING:
+            self._set_active_status()
```
dragomirp, line 1349: We should most probably don't change the status in `_set_active_status()` if the message is `INSUFFICIENT_SIZE_WARNING` and then reset the status here before running `_set_active_status()`.

This is because if another hook runs `_set_active_status()` the status will be overwritten unite the next update status. @marceloneppel what do you think?

lucasgameiroborges: I'm not sure I understand your point. Could you elaborate? My motivation for this check is to resolve the blocked status in case the storage space got released (to >10%)

lucasgameiroborges: the check for `INSUFFICIENT_SIZE_WARNING` is just to make sure that, if the charm was blocked due to insufficient size but the user has released enough space, then resolve. Maybe there's a better way?

marceloneppel: I think it's a good idea, @dragomirp.

@lucasgameiroborges, the idea is to just set something else in the status at this point, like the Active Status, and put a condition in the `_set_active_status` method not to change the status to Active if the current status is `INSUFFICIENT_SIZE_WARNING`. This way, we avoid some other places overring this status incorrectly (maybe [here](https://github.com/canonical/postgresql-k8s-operator/blob/a3f363a383da6969390ab3579343f1390aa43acd/src/charm.py#L940)).

I see that this will need to be handled correctly later by improving how we manage the unit's statuses across the entire charm code.

lucasgameiroborges: I get the reasoning for including a check inside `_set_active_status` so the blocked status doesn't get overwritten in other places, as for `just set something else in the status at this point, like the Active Status` do you mean using `self.unit.status = ActiveStatus()` directly? why?

dragomirp: The elif should be something like:
```
        elif self.unit.status.message == INSUFFICIENT_SIZE_WARNING:
            self.unit.status = ActiveStatus()
            self._set_active_status()
```
if `_set_active_status()` preserves `INSUFFICIENT_SIZE_WARNING` it should be unset before calling it. Otherwise it won't unset.

taurus-forever: Q: are we moving the right way here? 
It smells like Status/Message manipulation is going outside of our control.

lucasgameiroborges: Oh! Now I see the problem! Yes of course, thanks @dragomirp and @marceloneppel !

@taurus-forever : I agree that status management is a bit all over the place. AFAIK this is a problem in other charms too. Interesting discussion here: https://chat.canonical.com/canonical/pl/ncbsrryk378jux5r8cxjjngzna

dragomirp (approved):

marceloneppel (approved):

PR #688, opened 2024-09-09
[DPE-5116] Add pgAudit

marceloneppel:

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -303,6 +303,10 @@ options:
     default: false
     type: boolean
     description: Enable timescaledb extension
+  plugin_audit_enable:
+    default: false
+    type: boolean
+    description: Enable pgAudit extension
   profile:
    description: |
       Profile representing the scope of deployment, and used to tune resource allocation.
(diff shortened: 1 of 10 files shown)


config.yaml
```diff
@@ -303,6 +303,10 @@ options:
     default: false
     type: boolean
     description: Enable timescaledb extension
+  plugin_audit_enable:
+    default: false
```
taurus-forever, line 307: The same as VM, check default with Mohamed, please.

taurus-forever (approved):

lucasgameiroborges (approved):
Nice :)

dragomirp (approved):

config.yaml
marceloneppel: Thanks for pointing that out, Alex! After syncing with Mohamed, I updated the plugin to enable it by default on https://github.com/canonical/postgresql-k8s-operator/pull/688/commits/32437181df919b938509ed6e6081ee54e2f1f59c.

Issue #622, opened 2024-08-10
Successful restore does not trigger all the related events


Zvirovyi:
## Steps to reproduce

1. Deploy postgresql-k8s with s3-integrator
2. Create full backup (`juju run postgresql-k8s/leader create-backup`)
3. Restore it (`juju run postgresql-k8s/leader restore backup-id=<your_value>`)

## Expected behavior
After successful restore, cluster should print message "Restore succeeded" in the debug-log and Patroni on-failure condition should be restored.

## Actual behavior

Cluster appears in the "Move restored cluster to another S3 bucket" and it indicates, that restore process is finished. But there is no "Restore succeeded" message in the debug-log and Patroni on-failure condition isn't really restored - `juju show-unit postgresql-k8s/0` shows related fields:
```
overridden-patroni-on-failure-condition-repeat-cause: restore-backup
patroni-on-failure-condition-override: ignore
```
Overall, it leads to the conclusion, that successful restore does not trigger all its events and does not restore cluster to the fully working state.

## Versions

Operating system: Ubuntu 24.04 LTS

Juju CLI: 3.5.3-genericlinux-amd64

Juju agent: 3.5.3

Charm revision: 349

microk8s: MicroK8s v1.30.1 revision 6843

## Log output

Juju debug log: [k8s_is_restore_finished_backupid.log](https://github.com/user-attachments/files/16570895/k8s_is_restore_finished_backupid.log)

## Additional context

Although I can surely tell what commit exactly caused this issue in the VM version (see reference at the top), here [PITR](https://github.com/canonical/postgresql-k8s-operator/pull/554) was implemented with these changes all from the beginning. I think this issue will be critical when working on Timelines management for the PITR.

Zvirovyi:
#690 fixed this issue

PR #690, opened 2024-09-09
[MISC] Fixes for PITR backup test instabilities

lucasgameiroborges:
- This PR addresses some of the sources of instability in CI, mainly those introduced with the new PITR feature. 
- Unit testing will be implemented once you are OK with the changes.
- See review comments for specific context.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -337,8 +337,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         if scope not in get_args(Scopes):
             raise RuntimeError("Unknown secret scope.")
 
-        peers = self.model.get_relation(PEER)
-        if not peers:
+        if not (peers := self.model.get_relation(PEER)):
             return None
 
         secret_key = self._translate_field_to_secret_key(key)
@@ -356,8 +355,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         if not value:
             return self.remove_secret(scope, key)
 
-        peers = self.model.get_relation(PEER)
-        if not peers:
+        if not (peers := self.model.get_relation(PEER)):
             return None
 
         secret_key = self._translate_field_to_secret_key(key)
@@ -370,8 +368,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         if scope not in get_args(Scopes):
             raise RuntimeError("Unknown secret scope.")
 
-        peers = self.model.get_relation(PEER)
-        if not peers:
+        if not (peers := self.model.get_relation(PEER)):
             return None
 
         secret_key = self._translate_field_to_secret_key(key)
--- a/src/charm.py
+++ b/src/charm.py
@@ -10,7 +10,6 @@ import logging
 import os
 import re
 import sys
-import time
 from pathlib import Path
 from typing import Dict, List, Literal, Optional, Tuple, get_args
 
@@ -2092,18 +2091,13 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             )
         except ExecError:  # For Juju 2.
             logger.debug("Pebble logs command failed. Trying patroni logs directly")
-
-        counter = 0
-        while len(patroni_exceptions) == 0 and counter < 10:
-            time.sleep(3)
-            log_exec = container.pebble.exec(["cat", "/var/log/postgresql/patroni.log"])
+            log_exec = container.pebble.exec(["cat", "/var/log/postgresql/patroni.log*"])
             patroni_logs = log_exec.wait_output()[0]
             patroni_exceptions = re.findall(
                 r"^([0-9- :]+) UTC \[[0-9]+\]: INFO: removing initialize key after failed attempt to bootstrap the cluster",
                 patroni_logs,
                 re.MULTILINE,
             )
-            counter += 1
 
         if len(patroni_exceptions) > 0:
             logger.debug("Failures to bootstrap cluster detected on Patroni service logs")
--- a/src/charm.py
+++ b/src/charm.py
@@ -2091,13 +2091,32 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             )
         except ExecError:  # For Juju 2.
             logger.debug("Pebble logs command failed. Trying patroni logs directly")
-            log_exec = container.pebble.exec(["cat", "/var/log/postgresql/patroni.log*"])
+            log_exec = container.pebble.exec(["cat", "/var/log/postgresql/patroni.log"])
             patroni_logs = log_exec.wait_output()[0]
             patroni_exceptions = re.findall(
                 r"^([0-9- :]+) UTC \[[0-9]+\]: INFO: removing initialize key after failed attempt to bootstrap the cluster",
                 patroni_logs,
                 re.MULTILINE,
             )
+            # look for older logs
+            if len(patroni_exceptions) == 0:
+                logger.debug("Looking for older logs...")
+                log_exec = container.pebble.exec([
+                    "find",
+                    "/var/log/postgresql/",
+                    "-name",
+                    "'patroni.log.*'",
+                    "-exec",
+                    "cat",
+                    "{}",
+                    "+",
+                ])
+                patroni_logs = log_exec.wait_output()[0]
+                patroni_exceptions = re.findall(
+                    r"^([0-9- :]+) UTC \[[0-9]+\]: INFO: removing initialize key after failed attempt to bootstrap the cluster",
+                    patroni_logs,
+                    re.MULTILINE,
+                )
 
         if len(patroni_exceptions) > 0:
             logger.debug("Failures to bootstrap cluster detected on Patroni service logs")
--- a/src/charm.py
+++ b/src/charm.py
@@ -346,6 +346,9 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             return self.remove_secret(scope, key)
 
         peers = self.model.get_relation(PEER)
+        if not peers:
+            return None
+
         secret_key = self._translate_field_to_secret_key(key)
         # Old translation in databag is to be deleted
         self.peer_relation_data(scope).delete_relation_data(peers.id, [key])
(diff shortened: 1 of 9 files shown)


src/charm.py
```diff
@@ -549,11 +556,18 @@ def _on_peer_relation_changed(self, event: HookEvent) -> None:  # noqa: C901
...
+        services = container.pebble.get_services(names=[self._postgresql_service])
         if (
-            "restoring-backup" in self.app_peer_data or "restore-to-time" in self.app_peer_data
-        ) and isinstance(self.unit.status, BlockedStatus):
-            event.defer()
+            ("restoring-backup" in self.app_peer_data or "restore-to-time" in self.app_peer_data)
+            and len(services) > 0
+            and not self._was_restore_successful(container, services[0])
+        ):
+            logger.debug("on_peer_relation_changed early exit: Backup restore check failed")
```
lucasgameiroborges, line 570: It was observed in some cases that, after a PITR restore failed, the first event to run following the failure would be `on_peer_relation_changed` instead of `update_status` (where backup failures are normally spotted by the charm). When that happened, the charm would try to ping patroni, fail, enter `awaiting for member to start` status and endlessly defer, never knowing it was a backup failure. this condition is meant to avoid that.

src/charm.py
```diff
@@ -1338,6 +1352,23 @@ def _on_update_status_early_exit_checks(self, container) -> bool:
...
+        if self._has_blocked_status or self._has_non_restore_waiting_status:
+            # If charm was failing to disable plugin, try again and continue (user may have removed the objects)
+            if self.unit.status.message == EXTENSION_OBJECT_MESSAGE:
+                self.enable_disable_extensions()
+                return True
+
+            if (
+                self.unit.status.message == MOVE_RESTORED_CLUSTER_TO_ANOTHER_BUCKET
+                and "require-change-bucket-after-restore" not in self.app_peer_data
+            ):
```
lucasgameiroborges, line 1367: this extra condition is meant to allow `MOVE_RESTORED_CLUSTER_TO_ANOTHER_BUCKET` blocked status to resolve inside `update_status` event, in case the flag is no longer there, avoiding potential endless block.

src/charm.py
```diff
@@ -1413,23 +1435,25 @@ def _on_update_status(self, _) -> None:
...
-        if service.current != ServiceStatus.ACTIVE:
-            if "restore-to-time" in self.app_peer_data and all(self.is_pitr_failed(container)):
-                logger.error(
-                    "Restore failed: database service failed to reach point-in-time-recovery target. "
-                    "You can launch another restore with different parameters"
-                )
-                self.log_pitr_last_transaction_time()
-                self.unit.status = BlockedStatus(CANNOT_RESTORE_PITR)
-                return False
+        if "restore-to-time" in self.app_peer_data and all(self.is_pitr_failed(container)):
```
lucasgameiroborges, line 1438: often times, in juju 3 specifically, the restore would fail (making patroni unresponsive) but the service status itself would stay active for a while. This was causing the charm to think it was just patroni not being ready on the unit and never catching the underlying failure. Depending on the ordering of events following, this could cause the charm to go into `waiting`  status and never into `blocked`

src/charm.py
```diff
@@ -2091,26 +2115,56 @@ def is_pitr_failed(self, container: Container) -> Tuple[bool, bool]:
...
-            log_exec = container.pebble.exec(["cat", "/var/log/postgresql/patroni.log"])
-            patroni_logs = log_exec.wait_output()[0]
-            patroni_exceptions = re.findall(
-                r"^([0-9- :]+) UTC \[[0-9]+\]: INFO: removing initialize key after failed attempt to bootstrap the cluster",
-                patroni_logs,
-                re.MULTILINE,
-            )
+        patroni_exceptions = []
+        count = 0
+        while len(patroni_exceptions) == 0 and count < 10:
```
lucasgameiroborges, line 2120: This way of checking the logs for failures is not very stable and often doesn't catch the errors. My changes here aim to just decrease the likelihood of false negatives, but integration tests can still fail on 1st try, specially on juju 2.9. We probably need to redo this check using another approach.

tests/integration/test_backups.py
```diff
@@ -194,9 +194,15 @@ async def test_backup_aws(ops_test: OpsTest, cloud_configs: Tuple[Dict, Dict]) -
         await ops_test.model.wait_for_idle(status="active", timeout=1000)
 
     # Remove the database app.
-    await ops_test.model.remove_application(database_app_name, block_until_done=True)
+    await ops_test.model.remove_application(database_app_name)
```
lucasgameiroborges, line 197: this change (and the other similar ones) aim to avoid the case where, for some reason, the application removal gets stuck and never returns, causing the test to hang for 2h (until CI job itself times out). Because `remove_application()` does not have a timeout parameter on juju 2.9, I used the regular `block_until` with timeout.

tests/integration/test_backups.py
```diff
@@ -297,7 +309,7 @@ async def test_restore_on_new_cluster(ops_test: OpsTest, github_secrets) -> None
             database_app_name,
             0,
             S3_INTEGRATOR_APP_NAME,
-            MOVE_RESTORED_CLUSTER_TO_ANOTHER_BUCKET,
+            ANOTHER_CLUSTER_REPOSITORY_ERROR_MESSAGE,
```
lucasgameiroborges, line 312: Reverting the change made on PITR PR which was not fully appropriate (possibly motivated by the issue observed in https://github.com/canonical/postgresql-k8s-operator/pull/690/files#r1764218285). Double checked with @marceloneppel here.

tests/integration/new_relations/test_new_relations.py
```diff
@@ -91,7 +91,11 @@ async def test_database_relation_with_charm_libraries(ops_test: OpsTest, databas
             f"{APPLICATION_APP_NAME}:{FIRST_DATABASE_RELATION_NAME}", DATABASE_APP_NAME
         )
         await ops_test.model.wait_for_idle(
-            apps=[DATABASE_APP_NAME], status="active", raise_on_blocked=True, timeout=1000
+            apps=[DATABASE_APP_NAME],
+            status="active",
+            raise_on_blocked=True,
+            raise_on_error=False,
```
lucasgameiroborges, line 97: `raise_on_error=False` missing here, another one below.

tests/integration/helpers.py
```diff
@@ -744,7 +744,7 @@ async def switchover(
         candidate: The unit that should be elected the new primary.
     """
     primary_ip = await get_unit_address(ops_test, current_primary)
-    for attempt in Retrying(stop=stop_after_attempt(4), wait=wait_fixed(5), reraise=True):
+    for attempt in Retrying(stop=stop_after_attempt(60), wait=wait_fixed(3), reraise=True):
```
lucasgameiroborges, line 747: probably a bit overkill, but this was affecting `test_backups` which is an annoying test to retry and these values helped.

src/charm.py
```diff
@@ -1413,23 +1435,25 @@ def _on_update_status(self, _) -> None:
...
+            self.unit.status = BlockedStatus(CANNOT_RESTORE_PITR)
+            return False
 
-            if self.unit.status.message != CANNOT_RESTORE_PITR:
-                logger.error("Restore failed: database service failed to start")
-                self.unit.status = BlockedStatus("Failed to restore backup")
+        if (
+            service.current != ServiceStatus.ACTIVE
+            and self.unit.status.message != CANNOT_RESTORE_PITR
+        ):
```
marceloneppel, line 1450: I suspect that this block is being reached earlier than the one above on Juju 2.9 because of the status message (`Failed to restore backup`) of the unit in the failed tests:
* https://github.com/canonical/postgresql-k8s-operator/actions/runs/10911425853/job/30284387686#step:31:699
* https://github.com/canonical/postgresql-k8s-operator/actions/runs/10911425853/job/30284388001#step:31:214
As I commented before on MM, the fix for that may be investigated later in another PR.

marceloneppel (commented):
Amazing work, Lucas! Thanks for the improvements.

I left one request for a possible change in the way we check for a failure in PITR.

src/charm.py
```diff
@@ -2091,26 +2115,56 @@ def is_pitr_failed(self, container: Container) -> Tuple[bool, bool]:
...
+                )
+                if len(patroni_exceptions) != 0:
+                    break
+                # If no match, look at older logs
+                log_exec = container.pebble.exec([
+                    "find",
+                    "/var/log/postgresql/",
+                    "-name",
+                    "'patroni.log.*'",
+                    "-exec",
```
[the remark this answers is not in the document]
lucasgameiroborges: That's why we have [this check](https://github.com/canonical/postgresql-k8s-operator/blob/lucas/investigate-ci/src/charm.py#L2163-L2165) afterwards, isn't it? If we end up finding an older error from a previous restore attempt, this second return param will be `False` and [this check would fail](https://github.com/canonical/postgresql-k8s-operator/blob/lucas/investigate-ci/src/charm.py#L1438C56-L1438C91)

src/charm.py
lucasgameiroborges: That status is set [here after the PITR error check](https://github.com/canonical/postgresql-k8s-operator/blob/lucas/investigate-ci/src/charm.py#L1452), so I think the reason for this block running before the one above is because [this check failed when it shoud've succeeded](https://github.com/canonical/postgresql-k8s-operator/blob/lucas/investigate-ci/src/charm.py#L1438C56-L1438C91).

Anyway, I agree that this should be investigated in another PR, probably won't be a quick fix unfortunately.

src/charm.py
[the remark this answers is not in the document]
lucasgameiroborges: (BTW - I don't mind removing this if you want to be on the safe side, this check is not 100% resilient anyway)

tests/integration/helpers.py
lucasgameiroborges: In this PR commits there are 3 examples:
https://github.com/canonical/postgresql-k8s-operator/actions/runs/10894200338/job/30260414050#step:31:395

src/charm.py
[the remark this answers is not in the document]
marceloneppel: You're right. My mistake.

marceloneppel (approved):
LGTM! Thanks a lot, Lucas!

src/charm.py
```diff
@@ -356,6 +357,9 @@ def set_secret(self, scope: Scopes, key: str, value: Optional[str]) -> Optional[
             return self.remove_secret(scope, key)
 
         peers = self.model.get_relation(PEER)
+        if not peers:
+            return None
```
dragomirp, line 361:
```suggestion
        if not (peers := self.model.get_relation(PEER)):
            return None
```
You can do a walrus check to do the if and the assignment in the same go.

dragomirp (approved):

Zvirovyi:
Great findings and improvements! Actually, this PR will fix #622

taurus-forever (approved):
Wow. This is an excellent stabilization progress!

PR #693, opened 2024-09-11
[Hotfix] Remove failing tests from CI

lucasgameiroborges:
## Issue

Both `tests/integration/ha_tests/test_rollback_to_master_label.py` and `tests/integration/ha_tests/test_upgrade_to_primary_label.py` are failing and no longer needed since the new stable release, which is using the more appropriate `primary` label.

## Solution

~Remove them!~ Marking them as unstable for now.


marceloneppel (approved):

dragomirp (approved):

delgod (changes requested):
please mark tests unstable instead of removing if we need a hotfix asap.
Also, we still should support rollback to old revisions that used old labels

lucasgameiroborges:
> please mark tests unstable instead of removing if we need a hotfix asap.
Also, we still should support rollback to old revisions that used old labels

Quick context: Those 2 tests are very much a copy of the already existing `ha_tests/test_upgrade.py`, only using an old charm revision with `master` label instead of `primary`. When they were added in #532 it was said they should be removed once migration is done. Since the change made it to stable, it seemed to me we could stick with just `ha_tests/test_upgrade.py`.

Pinging @marceloneppel and @taurus-forever here before proceeding.

marceloneppel:
> please mark tests unstable instead of removing if we need a hotfix asap. Also, we still should support rollback to old revisions that used old labels

Hi, Mykola! As Lucas pointed out above, this migration logic should be removed (before the next stable with some more adjustments scheduled on https://warthogs.atlassian.net/browse/DPE-5443). The idea is to remove the unnecessary migration logic for the next upgrades to new stable revisions. We'll be using the `src/dependency.json` file with the `upgrade_supported` field to make users move from the previous stable to the current one before moving to the next stable that should have those tests and migration logic removed (making it safe not to have this tests and the migration code that will be removed in the attached Jira ticket).

Do you think there is a need to roll back from the next planned stable to another stable release before the current one?

taurus-forever (approved):
@delgod happy to discuss this.

PR #716, opened 2024-09-30
[DPE-5581] Timeline Management

### LINKED ISSUE #612: After scaling and changing s3-integrator config, "Move restored cluster to another S3 bucket" blocked status won't disappear for non-primary units
## Steps to reproduce

1. use 1 unit of postgresql-k8s with configured s3-integrator
2. restore backup
3. **single unit (restore can be done only on single unit) became in "Move restored cluster to another S3 bucket" blocked status as it should be**
4. `juju scale-application postgresql-k8s 2`
5. **both units became in "Move restored cluster to another S3 bucket" blocked status**
6. configure s3-integrator to use different path (for example, `juju config s3-integrator path=backup-test2`)
7. **observe, that only primary unit did loose "Move restored cluster to another S3 bucket" blocked status and became active / idle**
  Screenshot

## Expected behavior
All of the units should loose "Move restored cluster to another S3 bucket" blocked status after S3 configuration is changed.

## Actual behavior

See step 7. Also, it works for clusters with more than two units:
Screenshot

## Versions

Operating system: Ubuntu 24.04 LTS

Juju CLI: 3.5.3-genericlinux-amd64

Juju agent: 3.5.3

Charm revision: 347

microk8s: MicroK8s v1.30.1 revision 6843

## Log output

Juju debug log: [debug-log](https://github.com/user-attachments/files/16508345/issue_move_restored_cluster_scaling.log)

## Additional context

Basically, there is three paths to fix it:

1. Improve S3 relation code. So it will be related to the similar bug discovered previously - https://github.com/canonical/postgresql-k8s-operator/issues/591 and i think can be done in-batch.
2. Implement timelines management. Me, @taurus-forever and @marceloneppel tried to discuss it in the issue referenced above.
3. @taurus-forever proposed to block user from scaling application in potentially breaking scenarios.

Timelines management will:
  * allow us not to use Move restored cluster to another S3 bucket logic
  * allow user to see all the lifespan of cluster (not just separated to different S3 buckets after each restore) and effectively navigate through different timelines of it
  * imply WAL archiving refactor, that will fix issues like https://github.com/canonical/postgresql-k8s-operator/issues/614

### LINKED ISSUE #715: Add timeline management
This is continuation of the [point in time recovery work](https://github.com/canonical/postgresql-k8s-operator/pull/554). Currently, after each restore, cluster is required to be switched to the another s3 stanza because of lack of the timelines management. This feature should be implemented to improve user experience and allow user to restore through several timelines in the single s3 stanza.

Zvirovyi:
Add timeline management feature. Fixes #612 and fixes #715. [There are the same PR for the VM](https://github.com/canonical/postgresql-operator/pull/629) - basically, it's a port of the same code, so i left description unchanged, bet me know if it can be improved.
I've currently disabled failing unit tests and will fix them after this PR is reviewed.

# Overview
Main restriction of the [previous PITR work](https://github.com/canonical/postgresql-k8s-operator/pull/554) was lack of the WAL timeline management. Because of this, "move restored cluster to another s3 bucket" blocking message was introduced - to restrict user from having multiple timelines in the single s3 stanza. This PR is a continuation of that work and adds ability to have multiple WAL timelines in the single s3 stanza simultaneously and interact with them through UI.

# Changes
1. list-backups action:
1.1. `type` column renamed to the `action`; for the ordinary backups ` backup` suffix is added (example: `full` -> `full backup`)
1.2. restore events are shown along the ordinary backups to indicate timeline switch after each successful restore; user can use their id to run PITR restore within specific timeline chosen
1.3. `timeline` column added to indicate timeline of the backup / restore
  **Example can be found below in the section "Use-case example".**
2. `move restored cluster to another s3 bucket` message is removed entirely
3. check of the last archived WAL by the cluster compared to the one in the s3 stanza is removed entirely
4. ~`backup-id` is mandatory for the action `restore` again as user can now select specific timeline to run PITR within~. UPD: user can use `backup-id` to choose specific timeline to restore from, but with single `restore-to-time` parameter charm will automatically deduce required timeline as specified in https://github.com/canonical/postgresql-k8s-operator/pull/716#discussion_r1797136178.
5. improved `Restore succeeded` message to include info about:
5.1. real backup used to restore
5.2. timeline (either of real backup or of restore event) chosen for the restore
5.3. current timeline after restore
  **Example: `unit-postgresql-0: 01:17:27 INFO unit.postgresql/0.juju-log Restored to latest from timeline 2. Currently tracking the newly created timeline 3.`**

# Use-case example
PostgreSQL and s3-integrator are considered to be configured and integrated already. Timeline id of the newly created cluster will always be `1`.
For this example, we will create test table right after stanza initialization:
```sql
create table asd(message text);
```
Then, full backup must be created - it will be point of origin for WAL logs and therefore PITR (maybe this would be a great idea to automatically create full backup right after s3 stanza creation?):
```bash
juju run postgresql-k8s/leader create-backup
```
Then, let's write some test data and switch wal file (this ensures immediate wal file archiving and must be performed every time you want to preserve latest changes in the s3 stanza):
```sql
insert into asd values ('hello');
select pg_switch_wal();
```
Then, let's do restore to the full backup we created previously:
```bash
juju run postgresql-k8s/leader restore backup-id=2024-09-26T22:08:16Z
```
After cluster is restored successfully to the timeline 2, there should not be any data in the table. Let's create some test data specific to the timeline 2:
```sql
insert into asd values ('world');
select pg_switch_wal();
```
And then repeat restore to the full backup (on the timeline 1), but now with the `restore-to-time=latest` parameter:
```bash
juju run postgresql-k8s/leader restore backup-id=2024-09-26T22:08:16Z restore-to-time=latest
```
Now, in the newly created timeline 3, we will only see `hello` test data as we restored all the lifespan of the timeline 1 and not used timeline 2. Let's then restore to the timeline 2 using backup-id from the list-backups option:
image
```bash
juju run postgresql-k8s/leader restore backup-id=2024-09-26T22:11:54Z restore-to-time=latest
```
After successful restore to the timeline 4, we will see only the `world` from the timeline 2 in the test table and not the `hello` from the end of timeline 1.
Here the basic algorithm for this example:
```
1: table -> backup -> insert_hello -> restore_backup
2: insert_world -> restore_backup_latest
3: restore_timeline_2_latest
4: (*-*)
```

# PITR test algorithm
image

### DIFF
--- a/lib/charms/postgresql_k8s/v0/postgresql.py
+++ b/lib/charms/postgresql_k8s/v0/postgresql.py
@@ -36,7 +36,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 35
+LIBPATCH = 36
 
 INVALID_EXTRA_USER_ROLE_BLOCKING_MESSAGE = "invalid role(s) for extra user roles"
 
@@ -80,6 +80,10 @@ class PostgreSQLEnableDisableExtensionError(Exception):
 
 
 class PostgreSQLGetLastArchivedWALError(Exception):
+    """Exception raised when retrieving last archived WAL fails."""
+
+
+class PostgreSQLGetCurrentTimelineError(Exception):
     """Exception raised when retrieving last archived WAL fails."""
 
 
@@ -418,6 +422,16 @@ WHERE lomowner = (SELECT oid FROM pg_roles WHERE rolname = '{}');""".format(user
         except psycopg2.Error as e:
             logger.error(f"Failed to get PostgreSQL last archived WAL: {e}")
             raise PostgreSQLGetLastArchivedWALError()
+
+    def get_current_timeline(self) -> str:
+        """Get the timeline id for the current PostgreSQL unit."""
+        try:
+            with self._connect_to_database() as connection, connection.cursor() as cursor:
+                cursor.execute("SELECT timeline_id FROM pg_control_checkpoint();")
+                return cursor.fetchone()[0]
+        except psycopg2.Error as e:
+            logger.error(f"Failed to get PostgreSQL current timeline id: {e}")
+            raise PostgreSQLGetCurrentTimelineError()
 
     def get_postgresql_text_search_configs(self) -> Set[str]:
         """Returns the PostgreSQL available text search configs.
--- a/lib/charms/postgresql_k8s/v0/postgresql.py
+++ b/lib/charms/postgresql_k8s/v0/postgresql.py
@@ -84,7 +84,7 @@ class PostgreSQLGetLastArchivedWALError(Exception):
 
 
 class PostgreSQLGetCurrentTimelineError(Exception):
-    """Exception raised when retrieving last archived WAL fails."""
+    """Exception raised when retrieving current timeline id for the PostgreSQL unit fails."""
 
 
 class PostgreSQLGetPostgreSQLVersionError(Exception):
--- a/src/charm.py
+++ b/src/charm.py
@@ -82,7 +82,7 @@ from ops.pebble import (
 from requests import ConnectionError
 from tenacity import RetryError, Retrying, stop_after_attempt, stop_after_delay, wait_fixed
 
-from backups import CANNOT_RESTORE_PITR, MOVE_RESTORED_CLUSTER_TO_ANOTHER_BUCKET, PostgreSQLBackups
+from backups import CANNOT_RESTORE_PITR, PostgreSQLBackups
 from config import CharmConfig
 from constants import (
     APP_SCOPE,
@@ -966,15 +966,6 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         # The charm should not override this status outside of the function checking disk space.
         if self.unit.status.message == INSUFFICIENT_SIZE_WARNING:
             return
-        if "require-change-bucket-after-restore" in self.app_peer_data:
-            if self.unit.is_leader():
-                self.app_peer_data.update({
-                    "restoring-backup": "",
-                    "restore-stanza": "",
-                    "restore-to-time": "",
-                })
-            self.unit.status = BlockedStatus(MOVE_RESTORED_CLUSTER_TO_ANOTHER_BUCKET)
-            return
         try:
             if self._patroni.get_primary(unit_name_pattern=True) == self.unit.name:
                 self.unit.status = ActiveStatus("Primary")
@@ -1364,12 +1355,6 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
                 self.enable_disable_extensions()
                 return True
 
-            if (
-                self.unit.status.message == MOVE_RESTORED_CLUSTER_TO_ANOTHER_BUCKET
-                and "require-change-bucket-after-restore" not in self.app_peer_data
-            ):
-                return True
-
             logger.debug("on_update_status early exit: Unit is in Blocked/Waiting status")
             return False
         return True
@@ -1408,6 +1393,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
 
         if (
             "restoring-backup" not in self.app_peer_data
+            and "restore-to-time" not in self.app_peer_data
             and "stopped" not in self.unit_peer_data
             and services[0].current != ServiceStatus.ACTIVE
         ):
@@ -1459,15 +1445,33 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             logger.debug("Restore check early exit: Patroni has not started yet")
             return False
 
+        restoring_backup = self.app_peer_data.get("restoring-backup")
+        restore_timeline = self.app_peer_data.get("restore-timeline")
+        restore_to_time = self.app_peer_data.get("restore-to-time")
+        # Sometimes this check fails to connect to the PostgreSQL. Not happened on VM, so it's a k8s-specific try-catch
+        # try:
+        #     current_timeline = self.postgresql.get_current_timeline()
+        # except PostgreSQLGetCurrentTimelineError:
+        #     logger.debug("Restore check early exit: can't get current wal timeline")
+        current_timeline = self.postgresql.get_current_timeline()
+
         # Remove the restoring backup flag and the restore stanza name.
         self.app_peer_data.update({
             "restoring-backup": "",
             "restore-stanza": "",
             "restore-to-time": "",
+            "restore-timeline": "",
         })
         self.update_config()
         self.restore_patroni_on_failure_condition()
-        logger.info("Restore succeeded")
+
+        logger.info(
+            "Restored"
+            f"{' to ' + restore_to_time if restore_to_time else ''}"
+            f"{' from timeline ' + restore_timeline if restore_timeline and not restoring_backup else ''}"
+            f"{' from backup ' + self.backup._parse_backup_id(restoring_backup)[0] if restoring_backup else ''}"
+            f". Currently tracking the newly created timeline {current_timeline}."
+        )
 
         can_use_s3_repository, validation_message = self.backup.can_use_s3_repository()
         if not can_use_s3_repository:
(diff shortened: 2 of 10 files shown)


src/charm.py
```diff
@@ -1459,15 +1445,33 @@ def _was_restore_successful(self, container: Container, service: ServiceInfo) ->
...
             return False
 
+        restoring_backup = self.app_peer_data.get("restoring-backup")
+        restore_timeline = self.app_peer_data.get("restore-timeline")
+        restore_to_time = self.app_peer_data.get("restore-to-time")
+        # Sometimes this check fails to connect to the PostgreSQL. Not happened on VM, so it's a k8s-specific try-catch
+        # try:
+        #     current_timeline = self.postgresql.get_current_timeline()
+        # except PostgreSQLGetCurrentTimelineError:
+        #     logger.debug("Restore check early exit: can't get current wal timeline")
```
marceloneppel, line 1455: Did you manage to figure out and fix in this PR the reason why it was failing on K8s? If so, instead of commenting, you should remove this block of code.

src/charm.py
```diff
@@ -1459,15 +1445,33 @@ def _was_restore_successful(self, container: Container, service: ServiceInfo) ->
...
             "restore-to-time": "",
+            "restore-timeline": "",
         })
         self.update_config()
         self.restore_patroni_on_failure_condition()
-        logger.info("Restore succeeded")
+
+        logger.info(
+            "Restored"
+            f"{' to ' + restore_to_time if restore_to_time else ''}"
```
marceloneppel, line 1470: It would be a little bit weird to show `Restored to from timeline...`. Should the else clause return `latest` instead of an empty string?

src/charm.py
```diff
@@ -1459,15 +1445,33 @@ def _was_restore_successful(self, container: Container, service: ServiceInfo) ->
...
+            f"{' from timeline ' + restore_timeline if restore_timeline and not restoring_backup else ''}"
+            f"{' from backup ' + self.backup._parse_backup_id(restoring_backup)[0] if restoring_backup else ''}"
```
marceloneppel, line 1472: Also, here: will the charm reach this code without a backup ID?

tests/integration/test_backups.py
```diff
@@ -304,13 +294,7 @@ async def test_restore_on_new_cluster(ops_test: OpsTest, github_secrets) -> None
 
     # Wait for the restore to complete.
     async with ops_test.fast_forward():
-        await wait_for_idle_on_blocked(
-            ops_test,
-            database_app_name,
-            0,
-            S3_INTEGRATOR_APP_NAME,
-            ANOTHER_CLUSTER_REPOSITORY_ERROR_MESSAGE,
-        )
```
marceloneppel, line 313: I'd expect that the `ANOTHER_CLUSTER_REPOSITORY_ERROR_MESSAGE` blocked status would still be set if we tried to restore a backup to a different cluster. The new cluster could have a timeline with an equal number to the new timeline in the original cluster if we also perform a restore there.

templates/patroni.yml.j2
```diff
@@ -52,6 +52,7 @@ bootstrap:
     command: >
       pgbackrest --stanza={{ restore_stanza }} --pg1-path={{ storage_path }}/pgdata
       {%- if backup_id %} --set={{ backup_id }} {%- endif %}
+      {%- if restore_timeline %} --target-timeline="{{ restore_timeline }}" {% endif %}
```
marceloneppel, line 55: Doesn't `current` work as the value for the `target-timeline` parameter? If so, we could eliminate the need to pass a timeline as the backup ID in the restore action and match what was agreed upon in the specification document.

marceloneppel (commented):
Thanks for the amazing work, @Zvirovyi! I left some comments related to questions and suggestions.

src/charm.py
Zvirovyi: If user passed `latest` in the `restore_to_time` parameter - it will be in this variable (like `Restored to latest from backup ...` or `Restored to latest from timeline ...`). If it's an ordinary restore without PITR, it will entirely skip this line and just output `Restored from backup ...`.

src/charm.py
Zvirovyi: If user starts restore with `backup-id` pointing to the `action: restore`, so there wouldn't be a `restoring_backup` variable and only the `restore_timeline` with `restore_to_time`. But for the restore, backup-id is now required for user to specify.

templates/patroni.yml.j2
Zvirovyi: Do you mean the case when user uses real backup-id (with `action: full backup`, for example) and specifies restore-to-time? Or do you mean to eliminate ability to pass `action: restore` to the backup-id at the restore?

src/charm.py
Zvirovyi: I do not have this issue while testing manually, but in the integration tests this error occurred several times. I think it caused by #718 and safe to remove.

src/charm.py
marceloneppel: Got it! Thanks!

templates/patroni.yml.j2
marceloneppel: I mean letting the user be able to pass only the `restore-to-time` parameter (without requiring the user to pass the `backup-id` parameter too). I understand that the available timestamps we can restore to are unique among all the timelines because after we restore either to a backup-id or to a restore-to-time timestamp, the charm will generate a new timeline, so the new data will be written with the current datetime, not some timestamp from the past (that would be within the other timelines). With that, it seems that we'd be able to pass only the `restore-to-time` parameter to the restore action. Does that make sense?

I was wondering whether passing `current ` to the `target-timeline` parameter in the Patroni configuration file template would work to always restore to the right point in time, no matter what timeline it belongs to.

src/charm.py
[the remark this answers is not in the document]
Zvirovyi: I've changed it to use nested f-strings.

templates/patroni.yml.j2
Zvirovyi: `--target-timeline=current` will just set restoring timeline to the one that was active when chosen backup was created. And if there is no backup-id passed (`--set` parameter), PostgreSQL / pgBackRest will choose nearest backup before specified restore-to-time (`--target` parameter). And it will lead to very inconvenient UI as user will not always create backup after each restore. I think such functionality can take place when something like automatic backups creation system is introduced.

For example, in the PITR integration test algorithm only one single full backup used. In such situation user will be able to restore only along timeline 1.

I will attach additional diagram to visualize this issue.

image

tests/integration/test_backups.py
Zvirovyi: I've double checked this:
First of all, it should be safe to allow the new restored cluster to work with s3 stanza as after restore this cluster chooses the new timeline based on the stanza settings. For this particular example, the new restored cluster `new-postgresql-k8s` will operate on the timeline 4 as it supposed to be after 2 restores in the `test_backup_gcp` and the third one in the `test_restore_on_new_cluster` (1 -> 2 -> 3 -> 4).
Secondly, this blocked status emerges for the restored cluster due to the `stanza.get("name") != self.charm.app_peer_data.get("stanza", self.stanza_name)` check in the `can_use_s3_repository` func. Basically, it ensures the name of the juju model + application must be the same.
Because of the second point, I've restored this piece of code (this change was uncorrect at the first place) and now this test is correct.

templates/patroni.yml.j2
marceloneppel: Got it. Thanks for checking that. I'll just sync with Mohamed (our PR) about this requirement to see whether we need to change anything else.

marceloneppel: Copied from Matrix chat:

Do you think that the following approach would allow PITR with only the timestamp parameter (without requiring the user to pass the backup-id):
* Get the PITR timestamp.
* List backups by start time (timestamp) and restores by end time (timestamp).
* Find the nearest backup/restore with the timestamp before the PITR timestamp.
* Get the timeline ID of that backup/restore.
* Convert the timeline ID to hexadecimal (like mentioned at https://postgresqlco.nf/doc/en/param/recovery_target_timeline/).
* Pass the hexadecimal timeline ID to the --target-timeline parameter.

Vladyslav will test that later.

Zvirovyi: Added this functionality in both K8s and VM PRs.

marceloneppel (approved):
LGTM! Thanks for the great work @Zvirovyi! What needs to be added are the unit tests.

I'm currently reviewing the VM charm PR and running the integration tests locally to double-check that they're still working.

I asked the team to review this PR.

lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -83,6 +83,10 @@ class PostgreSQLGetLastArchivedWALError(Exception):
     """Exception raised when retrieving last archived WAL fails."""
 
 
+class PostgreSQLGetCurrentTimelineError(Exception):
+    """Exception raised when retrieving last archived WAL fails."""
```
lucasgameiroborges, line 87: Shouldn't the docstring be `Exception raised when retrieving current archived WAL fails.`?

Zvirovyi: Fixed :)

lucasgameiroborges: that was quick! thanks

lucasgameiroborges (approved):
LGTM! Nice work, and appreciate the detailed description!

taurus-forever (approved):
Thank you @Zvirovyi !!!

@marceloneppel please consider to merge this and update documentation as mentioned [here](https://github.com/canonical/postgresql-operator/pull/629#pullrequestreview-2368563187).

dragomirp (approved):

PR #733, opened 2024-10-09
[DPE-4469] Polish COS dashboard

lucasgameiroborges:
## Issue
The previous state of our COS dashboard (the pg one, didn't look at patroni one yet) has some inconsistencies and weirdnesses. This PR aims to fix most of them + renaming of the graphs, so it's clearer what metric they're showcasing.

The diff is very confusing - I'll put comments on the most important changes. Please try it out if you can, LMK your thoughts. Feel free to suggest any more metrics/graphs you would like to see.

## Screenshot Before
Screenshot from 2024-05-29 09-37-43

## Screenshot After
Screenshot 2024-10-10 at 15-29-37 PostgreSQL K8s - Dashboards - Grafana

### DIFF
--- a/src/grafana_dashboards/postgresql-metrics.json
+++ b/src/grafana_dashboards/postgresql-metrics.json
@@ -216,7 +216,8 @@
           "format": "time_series",
           "intervalFactor": 2,
           "legendFormat": "",
-          "refId": "A"
+          "refId": "A",
+          "type": "instant"
         }
       ],
       "thresholds": "",
@@ -558,7 +559,8 @@
           "expr": "pg_settings_max_connections{release=\"$release\", instance=\"$instance\"}",
           "format": "time_series",
           "intervalFactor": 1,
-          "refId": "A"
+          "refId": "A",
+          "type": "instant"
         }
       ],
       "thresholds": "",
@@ -620,7 +622,7 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "avg(rate(process_cpu_seconds_total{release=\"$release\", instance=\"$instance\"}[5m]) * 1000)",
+          "expr": "avg(irate(process_cpu_seconds_total{release=\"$release\", instance=\"$instance\"}[5m]) * 1000)",
           "format": "time_series",
           "intervalFactor": 2,
           "legendFormat": "CPU Time",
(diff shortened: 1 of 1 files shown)


src/grafana_dashboards/postgresql-metrics.json
```diff
@@ -216,7 +216,8 @@
           "format": "time_series",
           "intervalFactor": 2,
           "legendFormat": "",
-          "refId": "A"
+          "refId": "A",
+          "instant": true
```
lucasgameiroborges, line 220: all of the `"instant": true` additions are for the `singlestat` type panels: sometimes, if the user re-deployed the pg charm while monitoring, the value series get split in two (before and after). For graph panels it's mostly self-explanatory, but for single stats, the panel gets bloated with chars (it tries to print both values with the job names on it, see before screeshot of `Start Time` for an example). This way, we only fetch the instant current value.

src/grafana_dashboards/postgresql-metrics.json
```diff
@@ -620,7 +622,7 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "avg(rate(process_cpu_seconds_total{release=\"$release\", instance=\"$instance\"}[5m]) * 1000)",
+          "expr": "avg(irate(process_cpu_seconds_total{release=\"$release\", instance=\"$instance\"}[5m]) * 1000)",
```
lucasgameiroborges, line 625: `irate` allows for better visualizations of spikes in usage of both CPU time (here) and on Memory usage (below).

src/grafana_dashboards/postgresql-metrics.json
```diff
@@ -1995,7 +2006,7 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "pg_stat_database_tup_fetched{datname=~\"$datname\", instance=~\"$instance\"} != 0",
+          "expr": "irate(pg_stat_database_tup_fetched{datname=~\"$datname\", instance=~\"$instance\"}[5m])",
```
lucasgameiroborges, line 2009: for the rate-type metrics, we use `irate`. For count-type, we use the plain metric.

src/grafana_dashboards/postgresql-metrics.json
```diff
@@ -2191,10 +2200,10 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "pg_locks_count{datname=~\"$datname\", instance=~\"$instance\", mode=~\"$mode\"} != 0",
+          "expr": "irate(pg_stat_database_tup_updated{datname=~\"$datname\", instance=~\"$instance\"}[5m])",
```
lucasgameiroborges, line 2203: this kind of weird diffs are mostly repositioning of panels, they're not really being replaced.

src/grafana_dashboards/postgresql-metrics.json
```diff
@@ -2092,7 +2103,7 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "pg_stat_database_tup_inserted{datname=~\"$datname\", instance=~\"$instance\"} != 0",
+          "expr": "irate(pg_stat_database_tup_returned{datname=~\"$datname\", instance=~\"$instance\"}[5m])",
```
lucasgameiroborges, line 2106: to represent the number/rate of `SELECT`-type operations, I opted to use `tup_returned` instead of `tup_fetched` (which is still being shown on the graph, just under a different panel named ` read count/rate (fetched)`. They are slightly different metrics, theres a good explanation here: https://www.datadoghq.com/blog/postgresql-monitoring/#read-query-throughput-and-performance

taurus-forever (approved):
Thank you! The new data is now way better readable due to irate used.
The graphs orientation could create troubles on the dashboard refresh from upstream, but let's see the price here.

Good job! Thank you!

marceloneppel (approved):

Issue #747, opened 2024-10-17
Connection rejection rule in `pg_hba.conf`


marceloneppel:
## Steps to reproduce

1. Run backups periodically, every 2 hours.

## Expected behavior
No connection rejection rule:
```sh
root@postgresql-k8s-0:/# cat /var/lib/postgresql/data/pgdata/pg_hba.conf
Do not edit this file manually!
# It will be overwritten by Patroni!
local all backup peer map=operator
local all monitoring password
hostssl all all 0.0.0.0/0 md5
hostssl replication replication 127.0.0.1/32 md5
hostssl replication replication 127.0.0.6/32 md5
hostssl replication replication postgresql-k8s-0.postgresql-k8s-endpoints.prod-temporal.svc.cluster.local md5
hostssl replication replication postgresql-k8s-1.postgresql-k8s-endpoints.prod-temporal.svc.cluster.local md5
hostssl replication replication postgresql-k8s-2.postgresql-k8s-endpoints.prod-temporal.svc.cluster.local md5
EOF
```

## Actual behavior

Connection rejection rule:
```sh
root@postgresql-k8s-0:/# cat /var/lib/postgresql/data/pgdata/pg_hba.conf
# Do not edit this file manually!
# It will be overwritten by Patroni!
local all backup peer map=operator
local all monitoring password
hostssl all all 0.0.0.0/0 reject
hostssl all all postgresql-k8s-0.postgresql-k8s-endpoints.stg-data-visualization-34.svc.cluster.local md5
hostssl replication replication 127.0.0.1/32 md5
hostssl replication replication 127.0.0.6/32 md5
hostssl replication replication postgresql-k8s-0.postgresql-k8s-endpoints.stg-data-visualization-34.svc.cluster.local md5
hostssl replication replication postgresql-k8s-1.postgresql-k8s-endpoints.stg-data-visualization-34.svc.cluster.local md5
hostssl replication replication postgresql-k8s-2.postgresql-k8s-endpoints.stg-data-visualization-34.svc.cluster.local md5
```

## Versions

Juju agent: 3.4.5

Charm revision: 281

Environment: PS6

## Log output

Juju debug log: 

```sh
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/patroni/api.py", line 1384, in query
    with self.patroni.postgresql.connection().cursor() as cursor:
  File "/usr/lib/python3/dist-packages/patroni/postgresql/__init__.py", line 335, in connection
[...]

PR #751, opened 2024-10-24
[DPE-5689] Fix connection rejection rule in `pg_hba.conf`

lucasgameiroborges:
## Issue

See #747. When the `create-backup` action is run on a non-primary unit (the case when TLS is being used), the unit disables the connectivity to the cluster in order to prevent client operations mid-backup. Afterwards, this same unit should re-establish the connectivity. 

Problem: if a switchover happens mid-operation and the replica unit creating the backup becomes the new primary, it will not re-enable connectivity to the cluster due to not being a replica anymore.

## Solution

Do not check if the unit is the primary or not before re-enabling connectivity (there's no need), but instead checks if it disabled it before -- If so, then enables it back.

### DIFF
--- a/tests/unit/test_backups.py
+++ b/tests/unit/test_backups.py
@@ -1351,6 +1351,10 @@ Juju Version: test-juju-version
             ),
         ])
         assert _change_connectivity_to_database.call_count == 2
+        _change_connectivity_to_database.assert_has_calls([
+            call(connectivity=False),
+            call(connectivity=True),
+        ])
         mock_event.fail.assert_not_called()
         mock_event.set_results.assert_called_once_with({"backup-status": "backup created"})
 
--- a/tests/unit/test_backups.py
+++ b/tests/unit/test_backups.py
@@ -1332,6 +1332,28 @@ Juju Version: test-juju-version
         mock_event.fail.assert_not_called()
         mock_event.set_results.assert_called_once_with({"backup-status": "backup created"})
 
+        # Test when this unit is a replica but gets promoted to primary mid-way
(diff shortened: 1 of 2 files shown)


tests/unit/test_backups.py
```diff
@@ -1332,6 +1332,28 @@ def test_on_create_backup_action(harness):
...
+                f"backup/{harness.charm.model.name}.{harness.charm.cluster_name}/latest",
+                mock_s3_parameters,
+            ),
+            call(
+                "Stdout:\nfake stdout\n\nStderr:\nfake stderr\n",
+                f"backup/{harness.charm.model.name}.{harness.charm.cluster_name}/2023-01-01T09:00:00Z/backup.log",
+                mock_s3_parameters,
+            ),
+        ])
+        assert _change_connectivity_to_database.call_count == 2
```
marceloneppel, line 1353: Just to be safe about the parameters used in the calls.
```suggestion
        assert _change_connectivity_to_database.call_count == 2
        _change_connectivity_to_database.assert_has_calls([
            call(connectivity=False),
            call(connectivity=True),
        ])
```

marceloneppel (changes requested):
Great work, Lucas! I requested only a minor change in the test.

tests/unit/test_backups.py
lucasgameiroborges: done.

marceloneppel (approved):
LGTM! Thanks, Lucas!

dragomirp (approved):

PR #762, opened 2024-11-08
[TEST] Early exit on missing s3 rel

dragomirp:


taurus-forever:
@dragomirp , with this fix I do not see the leader_elected re-deferred events anylonger. Worth to merge edge, but maybe add debug message to the early exit, please.

marceloneppel (approved):

taurus-forever (approved):

PR #774, opened 2024-11-15
[DPE-6033] Preload shared libs on normal PG start

dragomirp:
## Issue
psycopg errors out after upgrade due to `shared_preload_libraries` directive missing in Patroni's `postgresql.parameters` config

## Solution
* Add shared_preload_libraries` to `postgresql.parameters` config
* Add resources to primary label upgrade tests

### DIFF
--- a/templates/patroni.yml.j2
+++ b/templates/patroni.yml.j2
@@ -115,6 +115,7 @@ postgresql:
   bin_dir: /usr/lib/postgresql/{{ version }}/bin
   listen: 0.0.0.0:5432
   parameters:
+    shared_preload_libraries: 'timescaledb,pgaudit'
     {%- if enable_pgbackrest_archiving %}
     archive_command: 'pgbackrest --stanza={{ stanza }} archive-push %p'
     {% else %}
--- a/tests/integration/ha_tests/test_rollback_to_master_label.py
+++ b/tests/integration/ha_tests/test_rollback_to_master_label.py
@@ -17,6 +17,7 @@ from ..helpers import (
     APPLICATION_NAME,
     CHARM_BASE,
     DATABASE_APP_NAME,
+    METADATA,
     get_leader_unit,
     get_primary,
     get_unit_by_index,
@@ -108,8 +109,11 @@ async def test_fail_and_rollback(ops_test, continuous_writes) -> None:
 
     application = ops_test.model.applications[DATABASE_APP_NAME]
 
+    resources = {"postgresql-image": METADATA["resources"]["postgresql-image"]["upstream-source"]}
+    application = ops_test.model.applications[DATABASE_APP_NAME]
+
     logger.info("Refresh the charm")
-    await application.refresh(path=fault_charm)
+    await application.refresh(path=fault_charm, resources=resources)
 
     logger.info("Get first upgrading unit")
     # Highest ordinal unit always the first to upgrade.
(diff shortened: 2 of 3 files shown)


templates/patroni.yml.j2
```diff
@@ -115,6 +115,7 @@ postgresql:
   bin_dir: /usr/lib/postgresql/{{ version }}/bin
   listen: 0.0.0.0:5432
   parameters:
+    shared_preload_libraries: 'timescaledb,pgaudit'
```
dragomirp, line 118: This was only added to `bootstrap.dcs.postgresql.parameters` (line 48) which AFAIK is only used during bootstrap of the cluster. I think we may be able to remove most of the params in `bootstrap.dcs.postgresql.parameters` but we'll need to investingate what is necessary to be there.

taurus-forever: @dragomirp please offload DPE-6033 to @marceloneppel . We also need CI test improvement. We have stable=>edge test, I am surprised to see this issue in my manual testing.

marceloneppel (approved):

templates/patroni.yml.j2
marceloneppel: After testing this PR, I saw that another error was thrown:
```sh
unit-pg-k8s-2: 17:28:32 ERROR unit.pg-k8s/2.juju-log failed to change plugins: 
  File "/var/lib/juju/agents/unit-pg-k8s-2/charm/lib/charms/postgresql_k8s/v0/postgresql.py", line 346, in enable_disable_extensions
    cursor.execute(
psycopg2.errors.UndefinedFile: could not open extension control file "/usr/share/postgresql/14/extension/pgaudit.control": No such file or directory


During handling of the above exception, another exception occurred:

  File "/var/lib/juju/agents/unit-pg-k8s-2/charm/./src/charm.py", line 719, in _handle_enable_disable_extensions
    self.postgresql.enable_disable_extensions(extensions, database)
  File "/var/lib/juju/agents/unit-pg-k8s-2/charm/lib/charms/tempo_coordinator_k8s/v0/charm_tracing.py", line 1064, in wrapped_function
    return callable(*args, **kwargs)  # type: ignore
  File "/var/lib/juju/agents/unit-pg-k8s-2/charm/lib/charms/postgresql_k8s/v0/postgresql.py", line 357, in enable_disable_extensions
    raise PostgreSQLEnableDisableExtensionError()
charms.postgresql_k8s.v0.postgresql.PostgreSQLEnableDisableExtensionError
unit-pg-k8s-2: 17:28:32 INFO unit.pg-k8s/2.juju-log pg-k8s/2 unit upgraded. Evaluate and run `resume-upgrade` action to continue upgrade
```

It happens because the upgraded replica is aware of the new plugins config options, but it tries to enable/disable them in the primary, which is still in the old charm revision.

We can fix it by changing the call to `self.enable_disable_extensions()` in the pebble-ready event handler to:
```python
        if self.upgrade.idle:
            self.enable_disable_extensions()
```

And test with:
```sh
juju deploy postgresql-k8s pg-k8s --trust -n 3 # Please, ensure that the primary is either unit 0 or unit 1.

juju run pg-k8s/leader pre-upgrade-check

juju refresh pg-k8s --channel 14/edge
```

taurus-forever (approved):

templates/patroni.yml.j2
taurus-forever: @dragomirp shared in MM: this conversation is no longer relevant. The issue for failing test was a wrong OCI resource (addressed here as well).

taurus-forever (approved):

PR #787, opened 2024-11-25
[DPE-6052] Allow `--restore-to-time=latest` without a `backup-id`

marceloneppel:
## Issue
When trying to restore to the latest time from a repository, the charm complains that a backup ID is required.

## Solution
Allow users to restore without a backup ID (with only the `--restore-to-time=latest` parameter).

Also, add a new check that prevents the PITR and warns the user that there is no base backup in the latest timeline if there isn't one. Previously, if the user tried to restore from that timeline, the restore would be reported as successful, but in fact, it didn't happen.

### DIFF
--- a/src/backups.py
+++ b/src/backups.py
@@ -1100,11 +1100,8 @@ Stderr:
             event.fail(validation_message)
             return False
 
-        if not event.params.get("backup-id") and event.params.get("restore-to-time") in (
-            None,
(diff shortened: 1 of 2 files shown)


taurus-forever (approved):
Thank you for the fast fix!

dragomirp (approved):

PR #795, opened 2024-11-29
[DPE-5827] Synchronous node count configuration

dragomirp:
Implements DA135 configuration for synchronous node count.

Adds `all` and  `majority` sync standby policies and ability to set constant number of sync standbys.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -441,10 +441,10 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         else:
             return None
 
-    def updated_synchronous_node_count(self, num_units: int | None = None) -> bool:
+    def updated_synchronous_node_count(self) -> bool:
         """Tries to update synchronous_node_count configuration and reports the result."""
         try:
-            self._patroni.update_synchronous_node_count(num_units)
+            self._patroni.update_synchronous_node_count()
             return True
         except RetryError:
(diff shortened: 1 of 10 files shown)


src/charm.py
```diff
@@ -441,10 +441,10 @@ def get_unit_ip(self, unit: Unit) -> Optional[str]:
         else:
             return None
 
-    def updated_synchronous_node_count(self, num_units: int | None = None) -> bool:
```
dragomirp, line 444: Not really passed anywhere.

tests/integration/ha_tests/test_synchronous_policy.py
```diff
@@ -0,0 +1,89 @@
...
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_build_and_deploy(ops_test: OpsTest) -> None:
+    """Build and deploy three unit of PostgreSQL."""
+    wait_for_apps = False
+    # It is possible for users to provide their own cluster for HA testing. Hence, check if there
+    # is a pre-existing cluster.
+    if not await app_name(ops_test):
+        wait_for_apps = True
+        await build_and_deploy(ops_test, 3, wait_for_idle=False)
```
dragomirp, line 20: A lot of the policies will be the same with 3 units. Should I try with 5?

marceloneppel (approved):
LGTM!

PR #862, opened 2025-02-17
[MISC] Define charm constants

sinclert-canonical:
This PR moves certain string into constant values, so that they could be referenced throughout the codebase.

These two in particular (i.e. the `admin` permission group and the `postgres` default database name) will be used in the upcoming LDAP support. These changes are proposed an isolation, as they have nothing to do with the LDAP support itself.

### DIFF
--- a/src/constants.py
+++ b/src/constants.py
@@ -3,6 +3,7 @@
 
 """File containing constants to be used in the charm."""
 
+DATABASE_DEFAULT_NAME = "postgres"
 DATABASE_PORT = "5432"
 PEER = "database-peers"
 BACKUP_USER = "backup"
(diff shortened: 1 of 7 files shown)


src/constants.py
```diff
@@ -3,6 +3,7 @@
 
 """File containing constants to be used in the charm."""
 
+DATABASE_NAME = "postgres"
```
dragomirp, line 6: As with the other PR, a more specific name should be better.

dragomirp (approved):

src/constants.py
sinclert-canonical: Renamed to `DATABASE_DEFAULT_NAME` in commit https://github.com/canonical/postgresql-k8s-operator/pull/862/commits/2e0c144a97fdea29c320ad52254394d5f6ee2b6a.

marceloneppel (approved):

PR #872, opened 2025-02-25
[MISC] Disable storage test on arm

dragomirp:
* Disable the storage test on arm
* Update libraries

### DIFF
--- a/tests/integration/test_storage.py
+++ b/tests/integration/test_storage.py
@@ -7,7 +7,7 @@ import logging
 import pytest
 from pytest_operator.plugin import OpsTest
 
-from .. import markers
+from . import markers
 from .helpers import (
     DATABASE_APP_NAME,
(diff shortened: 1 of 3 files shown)


tests/integration/test_storage.py
```diff
@@ -17,10 +18,10 @@
 
 logger = logging.getLogger(__name__)
 
-MAX_RETRIES = 20
 INSUFFICIENT_SIZE_WARNING = "<10% free space on pgdata volume."
 
 
+@markers.amd64_only
```
dragomirp, line 24: Hostpath  has no size limitations, so we are filling the host disk. We should most probably start using real volumes in the tests (Microceph? https://microk8s.io/docs/how-to-ceph)

taurus-forever (approved):

marceloneppel (approved):

PR #875, opened 2025-02-26
[DPE-6555] Add read only URIs

dragomirp:
- Implementing DA158: 
   - Always set a read only endpoint
   - Set read-only-uris if client has a recent data-platform-libs
- Update libs

dpl changes PR: https://github.com/canonical/data-platform-libs/pull/206


marceloneppel (approved):
LGTM!

taurus-forever (approved):

PR #883, opened 2025-03-06
[DPE-6344] LDAP II: Include charm libs

sinclert-canonical:
This is the 2nd PR to introduce LDAP support into PostgreSQL. The complete list of changes can be seen in [this branch](https://github.com/canonical/postgresql-k8s-operator/tree/sinclert/ldap-integration-all).

### Contents

This PR introduces the `ldap` V0 charm library, as well as the `certificate-transfer` V0 charm library. Despite having V1 of the latter already published, V0 requires `jsonschema` instead of `pydantic` which plays nicely with the version of the TLS charm lib we are using (see [code](https://github.com/canonical/postgresql-k8s-operator/blob/e1684df29ffb8fdc7a80e19847daab83e27cdf5b/lib/charms/tls_certificates_interface/v2/tls_certificates.py#L312)). I would say whenever we bump one, we bump the other.

### References

### DIFF
--- a/lib/charms/postgresql_k8s/v0/postgresql_tls.py
+++ b/lib/charms/postgresql_k8s/v0/postgresql_tls.py
@@ -194,15 +194,16 @@ class PostgreSQLTLS(Object):
 
     def _on_certificate_added(self, event: CertificateAddedEvent) -> None:
         """Enable TLS when TLS certificate is added."""
-        if not event.ca:
-            logger.debug("No certificate available.")
-            event.defer()
+        relation = self.charm.model.get_relation(TLS_TRANSFER_RELATION, event.relation_id)
+        if relation is None:
+            logger.error("Relationship not established anymore.")
             return
 
-        self.charm.set_secret(SCOPE, "ca", event.ca)
+        secret_name = f"ca-{relation.app.name}"
+        self.charm.set_secret(SCOPE, secret_name, event.ca)
 
         try:
-            if not self.charm.push_tls_files_to_workload():
+            if not self.charm.push_ca_file_into_workload(secret_name):
                 logger.debug("Cannot push TLS certificates at this moment")
                 event.defer()
                 return
@@ -213,11 +214,23 @@ class PostgreSQLTLS(Object):
 
     def _on_certificate_removed(self, event: CertificateRemovedEvent) -> None:
         """Disable TLS when TLS certificate is removed."""
-        self.charm.set_secret(SCOPE, "ca", None)
+        relation = self.charm.model.get_relation(TLS_TRANSFER_RELATION, event.relation_id)
+        if relation is None:
+            logger.error("Relationship not established anymore.")
+            return
 
-        if not self.charm.update_config():
-            logger.debug("Cannot update config at this moment")
+        secret_name = f"ca-{relation.app.name}"
+        self.charm.set_secret(SCOPE, secret_name, None)
+
+        try:
+            if not self.charm.clean_ca_file_from_workload(secret_name):
+                logger.debug("Cannot clean CA certificates at this moment")
+                event.defer()
+                return
+        except (PebbleConnectionError, PathError, ProtocolError, RetryError) as e:
+            logger.error("Cannot clean CA certificates: %r", e)
             event.defer()
+            return
 
     def _get_sans(self) -> dict:
         """Create a list of Subject Alternative Names for a PostgreSQL unit.
--- a/lib/charms/postgresql_k8s/v0/postgresql_tls.py
+++ b/lib/charms/postgresql_k8s/v0/postgresql_tls.py
@@ -6,8 +6,9 @@
 This class handles certificate request and renewal through
 the interaction with the TLS Certificates Operator.
 
-This library needs that https://charmhub.io/tls-certificates-interface/libraries/tls_certificates
-library is imported to work.
+This library needs that the following libraries are imported to work:
+- https://charmhub.io/certificate-transfer-interface/libraries/certificate_transfer
+- https://charmhub.io/tls-certificates-interface/libraries/tls_certificates
 
 It also needs the following methods in the charm class:
 — get_hostname_by_unit: to retrieve the DNS hostname of the unit.
@@ -24,6 +25,15 @@ import re
 import socket
 from typing import List, Optional
 
+from charms.certificate_transfer_interface.v0.certificate_transfer import (
+    CertificateAvailableEvent as CertificateAddedEvent,
+)
+from charms.certificate_transfer_interface.v0.certificate_transfer import (
+    CertificateRemovedEvent as CertificateRemovedEvent,
+)
+from charms.certificate_transfer_interface.v0.certificate_transfer import (
+    CertificateTransferRequires,
+)
 from charms.tls_certificates_interface.v2.tls_certificates import (
     CertificateAvailableEvent,
     CertificateExpiringEvent,
@@ -45,11 +55,12 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version.
-LIBPATCH = 13
+LIBPATCH = 14
 
 logger = logging.getLogger(__name__)
 SCOPE = "unit"
-TLS_RELATION = "certificates"
+TLS_CREATION_RELATION = "certificates"
+TLS_TRANSFER_RELATION = "send-ca-cert"
 
 
 class PostgreSQLTLS(Object):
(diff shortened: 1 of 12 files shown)


lib/charms/postgresql_k8s/v0/postgresql_tls.py
```diff
@@ -163,12 +186,39 @@ def _on_certificate_expiring(self, event: CertificateExpiringEvent) -> None:
...
+                event.defer()
+                return
+        except (PebbleConnectionError, PathError, ProtocolError, RetryError) as e:
+            logger.error("Cannot push TLS certificates: %r", e)
+            event.defer()
+            return
+
+    def _on_certificate_removed(self, event: CertificateRemovedEvent) -> None:
+        """Disable TLS when TLS certificate is removed."""
+        self.charm.set_secret(SCOPE, "ca", None)
```
marceloneppel, line 216: Can both the `certificates` and the `send-ca-cert` relations be established at the same time? If so, is there any chance that the `ca` field is empty while the `certificates` relation is still active after the CA is removed here?

sinclert-canonical: You are absolutely right: both the `certificates` and the `send-ca-cert` charm interfaces could be provided at the same time. Therefore, we would need to differentiate CA secrets in order to avoid overriding existing ones + avoid cleanup up the wrong ones.

Will add the relation ID / relation name as suffix to the secret name, as well as the container pushed file.

sinclert-canonical: See commit https://github.com/canonical/postgresql-k8s-operator/pull/883/commits/d01ef69ac7dac450d6ce77e6d4d5aeb8413987ca.

src/charm.py
```diff
@@ -1803,6 +1803,18 @@ def _peers(self) -> Relation:
         """
         return self.model.get_relation(PEER)
 
+    def _push_ca_file_to_workload(self, container: Container, ca: str, secret_name: str) -> None:
+        """Uploads CA certificate to the workload container."""
+        container.push(
+            f"/usr/local/share/ca-certificates/{secret_name}.crt",
```
dragomirp, line 1809: If we are factoring this out, might as well pass the location as well and use it for all the TLS files.

src/charm.py
```diff
@@ -1803,6 +1803,18 @@ def _peers(self) -> Relation:
...
+            ca,
+            make_dirs=True,
+            permissions=0o400,
+            user=WORKLOAD_OS_USER,
+            group=WORKLOAD_OS_GROUP,
+        )
+        container.exec(["update-ca-certificates"]).wait()
```
dragomirp, line 1816: IMHO we should keep this out of the function or keep the push code inline.

lib/charms/postgresql_k8s/v0/postgresql_tls.py
```diff
@@ -206,6 +269,10 @@ def is_ip_address(address: str) -> bool:
             "sans_dns": sans_dns,
         }
 
+    def get_ca_file(self, secret_name: str) -> Optional[str]:
+        """Prepare CA file for the certificate transfer case."""
+        return self.charm.get_secret(SCOPE, secret_name)
```
dragomirp, line 274: IMHO we shouldn't add pass through functions for a single call. I think it would be clearer to just call `get_secret` directly in `push_ca_file_into_workload`

dragomirp:
Please, bump the patch coverage to pass the codecov check.

src/charm.py
sinclert-canonical: Ok, let's do that instead.

lib/charms/postgresql_k8s/v0/postgresql_tls.py
sinclert-canonical: Yeah, I agree. I initially defined it here to mimic the nature of the `get_tls_files` [method](https://github.com/canonical/postgresql-k8s-operator/blob/1029fd9c852fc16c3d08ad003cdffc8e9ae80c6b/lib/charms/postgresql_k8s/v0/postgresql_tls.py#L209-L223), but I agree it feels weird to have the charm calling the PostgreSQL TLS lib, just for such lib to execute a charm method...

lib/charms/postgresql_k8s/v0/postgresql_tls.py
marceloneppel: Thanks!

sinclert-canonical:
🗞️  Just created tests to comply with codecov minimum coverage rules:

- For the PostgreSQL TLS lib: https://github.com/canonical/postgresql-k8s-operator/pull/883/commits/2a29db34ce445fd37fe9943a40f12734b854f9af.
- For the PostgreSQL charm: https://github.com/canonical/postgresql-k8s-operator/pull/883/commits/0845006bd24158674991be4f401aca30c6c63618.

lib/charms/postgresql_k8s/v0/postgresql_tls.py
```diff
@@ -24,6 +25,15 @@
...
 
+from charms.certificate_transfer_interface.v0.certificate_transfer import (
+    CertificateAvailableEvent as CertificateAddedEvent,
+)
+from charms.certificate_transfer_interface.v0.certificate_transfer import (
+    CertificateRemovedEvent as CertificateRemovedEvent,
+)
+from charms.certificate_transfer_interface.v0.certificate_transfer import (
+    CertificateTransferRequires,
+)
```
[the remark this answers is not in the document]
sinclert-canonical: Good point.

Worth clarifying: splitting the import into 3 different statements **is required by ruff**. I trying putting all three on the same statement, but ruff kept refactoring back. I think it is related with the aliases 🤷🏻‍♂️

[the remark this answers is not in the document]
sinclert-canonical: Good point.

Worth clarifying: splitting the import into 3 different statements **is required by ruff**. I tried putting all three on the same statement, but ruff kept refactoring back. I think it is related with the aliases 🤷🏻‍♂️

src/charm.py
```diff
@@ -1811,41 +1824,36 @@ def push_tls_files_to_workload(self, container: Container = None) -> bool:
...
+
+    def push_ca_file_into_workload(self, secret_name: str) -> bool:
+        """Uploads CA certificate into the workload container."""
+        container = self.unit.get_container("postgresql")
+        certificates = self.get_secret(UNIT_SCOPE, secret_name)
+
+        if certificates is not None:
+            self._push_file_to_workload(
+                container=container,
+                file_path=f"{self._certs_path}/{secret_name}.crt",
```
taurus-forever, line 1845: I am not sure here. @delgod can we drop CA into `/usr/local/share/ca-certificates`?

taurus-forever (commented):
LGTM overall, two questions inside.

metadata.yaml
```diff
@@ -63,10 +63,17 @@ requires:
     interface: tls-certificates
     limit: 1
     optional: true
+  send-ca-cert:
```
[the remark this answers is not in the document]
sinclert-canonical: Good catch! I thought both provider and requirer would need to define the **exact same relation name**, but judging by what other charms are doing, `receive-ca-cert` is a popular relation name. I think that as long as we  keep using the certificate_transfer [interface](https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/certificate_transfer/v0), we should be good to go.

Done in commit https://github.com/canonical/postgresql-k8s-operator/pull/883/commits/c9723e0863cc1e7c63b2fc729f0daaec4ea74a6d.

src/charm.py
sinclert-canonical: To be clear: we were already storing the `ca.crt` file coming out of the `self-signed-certificates` charm into this location. The proposed change generalizes the function in order to also store `ca-{app-name}.crt` files.

Please clarify whether this is expected.

taurus-forever (approved):
LGTM, please sign it with Mykola to avoid surprises with CA.

metadata.yaml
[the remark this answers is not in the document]
taurus-forever: Thank you! Just for the history (MongoDB spec comment):
the `send-ca-cert` and `receive-ca-cert` are endpoint names.

In this case, the direction of CA exchange is clear:
`juju integrate glauth:send-ca-cert postgresql:receive-ca-cert`
(this is necessary for the self-signed CA only IF different CAs used for IAM and DB models.
It is recommended to deploy separate TLS model and consume the same CA in IAM and DB models (c) TLS team).

PR #884, opened 2025-03-06
[MISC] Skip backup tests without creds

dragomirp:
## Issue
Forked PRs have no access to Github secrets, so cloud backup tests will always fail.

## Solution
* Update charm libs
* Split the cloud config fixture
* Skip backup tests if no credentials are provided

Tested on https://github.com/canonical/postgresql-k8s-operator/pull/885

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -70,9 +70,9 @@ def cleanup_cloud(config: dict[str, str], credentials: dict[str, str]) -> None:
 async def aws_cloud_configs(ops_test: OpsTest) -> None:
     if (
         "AWS_ACCESS_KEY" not in os.environ
-        or not os.envioron["AWS_ACCESS_KEY"].strip()
+        or not os.environ["AWS_ACCESS_KEY"].strip()
         or "AWS_SECRET_KEY" not in os.environ
-        or not os.envioron["AWS_SECRET_KEY"].strip()
+        or not os.environ["AWS_SECRET_KEY"].strip()
     ):
         pytest.skip("AWS configs not set")
         return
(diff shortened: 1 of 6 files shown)


tests/integration/test_backups_aws.py
```diff
@@ -41,60 +38,12 @@
 
 logger = logging.getLogger(__name__)
 
-AWS = "AWS"
-GCP = "GCP"
-
-
-@pytest.fixture(scope="module")
-async def cloud_configs(ops_test: OpsTest) -> None:
```
dragomirp, line 49: Moved to conftest, it's the same everywhere.

tests/integration/conftest.py
```diff
@@ -11,3 +23,78 @@ def charm():
...
+        "s3", endpoint_url=construct_endpoint(config["endpoint"], config["region"])
+    )
+    bucket = s3.Bucket(config["bucket"])
+    # GCS doesn't support batch delete operation, so delete the objects one by one.
+    for bucket_object in bucket.objects.filter(Prefix=config["path"].lstrip("/")):
+        bucket_object.delete()
+
+
+@pytest.fixture(scope="module")
+async def aws_cloud_configs(ops_test: OpsTest) -> None:
```
dragomirp, line 70: Split the fixture to per cloud. So that we can skip only for the cloud that has missing credentials.

tests/integration/conftest.py
```diff
@@ -11,3 +23,78 @@ def charm():
...
+    if (
+        "AWS_ACCESS_KEY" not in os.environ
+        or not os.environ["AWS_ACCESS_KEY"].strip()
+        or "AWS_SECRET_KEY" not in os.environ
+        or not os.environ["AWS_SECRET_KEY"].strip()
+    ):
+        pytest.skip("AWS configs not set")
+        return
```
dragomirp, line 78: If the envvars are not set empty skip the test.

PR #881, opened 2025-03-05
[DPE-6344] LDAP I: Create access groups

sinclert-canonical:
This is the 1st PR to introduce LDAP support into PostgreSQL. The complete list of changes can be seen in [this branch](https://github.com/canonical/postgresql-k8s-operator/tree/sinclert/ldap-integration-all).

### Contents

This PR defines three different _access groups_, in order to differentiate authentication methods within the `pg_hba.conf` file when LDAP is activated (i.e. the GLAuth charm related). These _access groups_ are not yet put into such file, but they are leveraged in order to:
- Assign any charm internal user to the _internal-access_ group.
- Assign any charm relation user to the _relation-access_ group.

### References

- PostgreSQL-LDAP [specification](https://docs.google.com/document/d/1Wt9VhdYCVzPh6qfwYiKOcfQwQ6UW6765S09dFjnuBXY/edit?tab=t.0).

PR #886, opened 2025-03-07
[DPE-6344] LDAP III: Define config and handlers

sinclert-canonical:
This is the 3rd PR to introduce LDAP support into PostgreSQL. The complete list of changes can be seen in [this branch](https://github.com/canonical/postgresql-k8s-operator/tree/sinclert/ldap-integration-all).

### Contents

This PR extends the charm configuration to include all the initially proposed configuration options, which are useful for local debugging (some of them will go away before marking the integration complete). In addition, it creates the [PostgreSQLLDAP](https://github.com/canonical/postgresql-k8s-operator/blob/ldap-integration/handler-methods/src/ldap.py) class to hold all the GLAuth related handlers, similar to how other classes within the charm hold all the handlers for a particular functionality (i.e. `PostgreSQLBackups`, `PostgreSQLUpgrades`, etc).

### References

---

Depends on PRs https://github.com/canonical/postgresql-k8s-operator/pull/881 and https://github.com/canonical/postgresql-k8s-operator/pull/883.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -69,29 +69,12 @@ options:
       Enable synchronized sequential scans.
     type: boolean
     default: true
-  ldap_base_dn:
-    description: |
-      Root DN to begin the search for the user in.
-    type: string
-  ldap_bind_dn:
-    description: |
-      DN of user to bind to the directory with to perform the search.
-    type: string
-  ldap_bind_password:
-    description: |
-      Password for user to bind to the directory with to perform the search.
-    type: string
   ldap_search_filter:
     description: |
       The LDAP search filter to match users with.
       Example: (|(uid=$username)(email=$username))
     type: string
     default: "(uid=$username)"
-  ldap_url:
-    description: |
-      An RFC 4516 LDAP URL. Combines the LDAP scheme, host and port.
-      Example: ldap://ldap.example.net:3893
-    type: string
   logging_client_min_messages:
     description: |
       Sets the message levels that are sent to the client.
--- a/config.yaml
+++ b/config.yaml
@@ -69,6 +69,29 @@ options:
       Enable synchronized sequential scans.
     type: boolean
     default: true
+  ldap_base_dn:
+    description: |
+      Root DN to begin the search for the user in.
+    type: string
+  ldap_bind_dn:
+    description: |
+      DN of user to bind to the directory with to perform the search.
+    type: string
+  ldap_bind_password:
+    description: |
+      Password for user to bind to the directory with to perform the search.
+    type: string
+  ldap_search_filter:
+    description: |
+      The LDAP search filter to match users with.
+      Example: (|(uid=$username)(email=$username))
+    type: string
+    default: "(uid=$username)"
+  ldap_url:
+    description: |
+      An RFC 4516 LDAP URL. Combines the LDAP scheme, host and port.
+      Example: ldap://ldap.example.net:3893
+    type: string
   logging_client_min_messages:
     description: |
       Sets the message levels that are sent to the client.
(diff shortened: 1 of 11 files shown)


sinclert-canonical:
PR branch rebased from `main`.

marceloneppel (commented):
Great work, Sinclert!

Maybe setting the base branch to one of the branches of the previous LDAP PRs may help the other reviewers to review this PR quickly (as the diff will be smaller) as right now this PR has changes from them.

src/charm.py
```diff
@@ -1226,22 +1235,29 @@ def _on_get_password(self, event: ActionEvent) -> None:
         If no user is provided, the password of the operator user is returned.
         """
         username = event.params.get("username", USER)
+        if username not in PASSWORD_USERS and self.is_ldap_enabled:
+            event.fail("The action cannot be performed when LDAP is enabled")
```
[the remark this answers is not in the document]
sinclert-canonical: That statement would not be entirely true, as that conditional may be true when the Juju cluster admin specifies a **relation** user, instead of a system / internal one. If you think the proposed message is not complete enough, we could say: _The action cannot be performed for non-system users when LDAP is enable_

WDYT?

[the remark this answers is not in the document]
marceloneppel: It looks better. I liked it.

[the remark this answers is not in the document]
sinclert-canonical: Done in commit https://github.com/canonical/postgresql-k8s-operator/pull/886/commits/7a945979140b355ee4778fcc72c22aa9559814ed.

marceloneppel (approved):

sinclert-canonical:
🗞️ PR branch just rebased from `main`, after PRs https://github.com/canonical/postgresql-k8s-operator/pull/881 and https://github.com/canonical/postgresql-k8s-operator/pull/883 got merged.

Review should be easier now.

config.yaml
```diff
@@ -69,6 +69,29 @@ options:
...
+    description: |
+      The LDAP search filter to match users with.
+      Example: (|(uid=$username)(email=$username))
+    type: string
+    default: "(uid=$username)"
+  ldap_url:
+    description: |
+      An RFC 4516 LDAP URL. Combines the LDAP scheme, host and port.
+      Example: ldap://ldap.example.net:3893
+    type: string
```
taurus-forever, line 94: The spec mentions ldap_map + ldap_search_filter only. Where is the truth?

taurus-forever (approved):
LGTM, one question about config options mismatch with the spec.
CC: @7annaba3l , Mohamed , please sign UI.

config.yaml
sinclert-canonical: Yes, those config options were removed from the spec. I left them here for debugging purposes **until the whole integration is completed**, as stated in the PR [description](https://github.com/canonical/postgresql-k8s-operator/pull/886).

If you think there is a risk users installing the PostgreSQL-K8s charm using `--channel edge` may start using them, please, let me know so we do not even expose them for debugging reasons.

sinclert-canonical: Let me invert the approach in order to speed things up: I will prune the debugging config options, and only re-state them if we need to debug, somehow.

See commit https://github.com/canonical/postgresql-k8s-operator/pull/886/commits/73edc0e3a647aab5888ab15be481e5d844df3d85.

sinclert-canonical: Let me invert the approach in order to speed things up: I will prune the debugging config options, and only re-state them if we need to debug, somehow.

See commit https://github.com/canonical/postgresql-k8s-operator/pull/886/commits/b9751c1b094d4a3bb64e934c928f87e10b64e5bb.

PR #892, opened 2025-03-18
[DPE-6675] Add wal_keep_size config option

marceloneppel:
## Issue
The `wal_keep_size` PostgreSQL parameter is not configurable through the charm.

## Solution
Add that parameter as a config option.

It needs to be set through Patroni dynamic configuration as per https://patroni.readthedocs.io/en/latest/patroni_configuration.html.


dragomirp (approved):

7annaba3l:
Looks good to me also @marceloneppel , @taurus-forever .

PR #897, opened 2025-03-21
[DPE-6344] LDAP IV: Define pebble service

sinclert-canonical:
This is the 4th PR to introduce LDAP support into PostgreSQL. The complete list of changes can be seen in [this branch](https://github.com/canonical/postgresql-k8s-operator/tree/sinclert/ldap-integration-all).

### Contents

This PR leverages the upcoming `ldap-sync` snap service in order to start the synchronization of LDAP users whenever the PostgreSQL-K8s charm relates to the GLAuth-K8s one. Given that the service has not yet defined within the snap, the method depending on such service existing has been temporarily disabled.

### References

---

Depends on PR https://github.com/canonical/charmed-postgresql-snap/pull/88.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1978,18 +1978,19 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
 
     def _restart_ldap_sync_service(self) -> None:
         """Restart the LDAP sync service in case any configuration changed."""
-        if not self.is_primary:
-            return
-
         container = self.unit.get_container("postgresql")
         sync_service = container.pebble.get_services(names=[self.ldap_sync_service])
 
-        if not self.is_ldap_enabled and sync_service[0].is_running():
+        if not self.is_primary and sync_service[0].is_running():
+            logger.debug("Stopping LDAP sync service. It must only run in the primary")
+            container.stop(self.pg_ldap_sync_service)
+
+        if self.is_primary and not self.is_ldap_enabled:
             logger.debug("Stopping LDAP sync service")
             container.stop(self.ldap_sync_service)
             return
 
-        if self.is_ldap_enabled:
+        if self.is_primary and self.is_ldap_enabled:
             container.add_layer(
                 self.ldap_sync_service,
                 Layer({"services": {self.ldap_sync_service: self._generate_ldap_service()}}),
--- a/src/charm.py
+++ b/src/charm.py
@@ -2090,7 +2090,11 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
 
         self._handle_postgresql_restart_need()
         self._restart_metrics_service()
-        self._restart_ldap_sync_service()
+
+        # TODO: Un-comment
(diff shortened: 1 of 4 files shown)


src/charm.py
```diff
@@ -1983,21 +2063,8 @@ def update_config(self, is_creating_backup: bool = False) -> bool:
...
-            f"user={MONITORING_USER} password={self.get_secret('app', MONITORING_PASSWORD_KEY)} "
-        ):
-            container.add_layer(
-                self._metrics_service,
-                Layer({"services": {self._metrics_service: self._generate_metrics_service()}}),
-                combine=True,
-            )
-            container.restart(self._metrics_service)
+        self._restart_metrics_service()
+        # self._restart_ldap_sync_service()
```
taurus-forever, line 2067: should it stay commented (here and in test below)?

taurus-forever (approved):
LGTM, one question.
P.S. `_postgresql_service` => `postgresql_service` is up to @marceloneppel .

src/charm.py
sinclert-canonical: When the whole integration is completed: no, it should not.

However, I wanted to de-couple the contents of this PR, from the existence of the `start-ldap-synchronizer.sh` shell script within the charmed-postgresql-snap, being proposed in [this PR](https://github.com/canonical/charmed-postgresql-snap/pull/88/files). The delay is due to the anachronistic way of automatically uploading Python packages to PPAs, as shared with you yesterday.

src/charm.py
```diff
@@ -1893,14 +1933,54 @@ def _restart(self, event: RunWithLock) -> None:
...
+                self.metrics_service,
+                Layer({"services": {self.metrics_service: self._generate_metrics_service()}}),
+                combine=True,
+            )
+            container.restart(self.metrics_service)
+
+    def _restart_ldap_sync_service(self) -> None:
+        """Restart the LDAP sync service in case any configuration changed."""
+        if not self.is_primary:
+            return
```
marceloneppel, line 1957: Can we have a situation where the service is started in the primary, then a switchover happens and another unit becomes the primary, so we have the service running in two units? If so, will be there any problem?

sinclert-canonical: That is a great point, and yes, I think it could happen. In those cases, we would like to spawn the background process in the `primary` instance, and immediately stop it in the replicas.

See commit https://github.com/canonical/postgresql-k8s-operator/pull/897/commits/1a8217361eb35615d34794ee33a4c9287600341a.

marceloneppel: Great. Thanks!

marceloneppel (approved):
`_postgresql_service` => `postgresql_service` LGTM.

src/charm.py
sinclert-canonical: Latest version of the package (version `0.3.0`) uploaded to [the PPA](https://launchpad.net/~data-platform/+archive/ubuntu/postgresql-ldap-sync). Let's merge that PR first, so that we can un-comment the use of the brand new snap service.

PR #919, opened 2025-04-08
[DPE-6899] User->databases pg_hba rules

marceloneppel:
## Issue
Database access is not restricted through the `pg_hba.conf` file.

## Solution
Retrieve the list that each user should have access to and rebuild the ` pg_hba.conf` file rules based on that list, while keeping access to all the databases for the system users and superusers.

For the users created by this charm, some calls to the `update_config` method were added to the client relations files (`src/relations/db.py` and `src/relations/postgresql_provider.py`) to rebuild the `pg_hba.conf` file rules.

For the users created by the PgBouncer charm, the PostgreSQL charm needs to gather that information in some way because the `data_interfaces` doesn't share it yet. For that, an event trigger is used. That should be removed after we improve the interface.

An observer process (like the one we have in the VM charm) was created to update the list of users->databases in the replicas because the event trigger is fired only in the primary.

An integration test (`tests/integration/test_pg_hba.py`) was created to validate users created by both this and an external charms (PgBouncer is simulated through creating users by using the Data Integration relation user).

Still in progress (while you review the PR): fixing the failing integration tests and increasing coverage.

## Checklist


lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -697,25 +750,91 @@ def set_up_database(self) -> None:
...
         try:
+            with self._connect_to_database(
+                database="template1"
+            ) as connection, connection.cursor() as cursor:
+                cursor.execute(
+                    "SELECT TRUE FROM pg_event_trigger WHERE evtname = 'update_pg_hba_on_create_schema';"
+                )
+                if cursor.fetchone() is None:
+                    cursor.execute("""
+CREATE OR REPLACE FUNCTION update_pg_hba()
```
marceloneppel, line 761: Function used to update the `pg_hba` file when PgBouncer creates or deletes an user.

scripts/authorisation_rules_observer.py
```diff
@@ -0,0 +1,92 @@
...
+        for member in cluster_status["members"]:
+            # Check if the current member is the primary.
+            if member["name"] == member_name:
+                is_primary = True
+                break
+
+        if is_primary:
+            # Read contents from the pg_hba.conf file.
+            with open("/var/lib/postgresql/data/pgdata/pg_hba.conf") as file:
+                current_authorisation_rules = file.read()
```
dragomirp, line 72: Going forward, we should most probably just hash the file.

taurus-forever (approved):
Thank you!

dragomirp (approved):

PR #932, opened 2025-04-21
[DPE-6344] Remove CA transferred check

sinclert-canonical:
This PR removes the `ca_transferred` property from the PostgreSQLLDAP class, so that relationship could be established cross-model, which is not possible at the moment. This change harmonizes the logic with respect to its PostgreSQL-VM operator equivalent.

### Additional changes
- Fixes a typo related to how the pebble service is called, when it needs to be stopped.


taurus-forever (approved):
Thank you for addressing this to sync with VM charm!

dragomirp (approved):

marceloneppel (approved):

PR #933, opened 2025-04-21
[MISC] Don't set tls flag if relation isn't initialised

dragomirp:
## Checklist

### DIFF
--- a/src/relations/postgresql_provider.py
+++ b/src/relations/postgresql_provider.py
@@ -265,8 +265,9 @@ class PostgreSQLProvider(Object):
             ca = ""
 
         for relation in relations:
-            self.database_provides.set_tls(relation.id, tls)
-            self.database_provides.set_tls_ca(relation.id, ca)
+            if self.database_provides.fetch_relation_field(relation.id, "database"):
+                self.database_provides.set_tls(relation.id, tls)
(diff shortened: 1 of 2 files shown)


src/relations/postgresql_provider.py
```diff
@@ -265,8 +265,9 @@ def update_tls_flag(self, tls: str) -> None:
             ca = ""
 
         for relation in relations:
-            self.database_provides.set_tls(relation.id, tls)
-            self.database_provides.set_tls_ca(relation.id, ca)
+            if self.database_provides.fetch_relation_field(relation.id, "database"):
+                self.database_provides.set_tls(relation.id, tls)
+                self.database_provides.set_tls_ca(relation.id, ca)
```
dragomirp, line 270: Check if the relation is already properly intialised. If not the values should be set by the `database_requested` hook.

tests/integration/helpers.py
```diff
@@ -108,7 +108,6 @@ async def build_and_deploy(
             apps=[database_app_name],
             status=status,
             raise_on_blocked=True,
-            raise_on_error=False,
```
dragomirp, line 111: We ignored this due to early volume mount errors, but it will also hide other errors during start up.

marceloneppel (approved):
Thanks!

PR #993, opened 2025-06-10
Fix broken URLs and spelling errors

a-velasco:
## Issue
The automatic doc checks did not pass.

## Solution
Updated all broken links, fixed spelling errors, and added workarounds for false positives.

Note: The spellcheck tool will be updated in the near future to one that gives us more freedom to define exceptions and custom rules with regex. 

## Checklist

### DIFF
--- a/docs/reference/system-requirements.md
+++ b/docs/reference/system-requirements.md
@@ -7,7 +7,7 @@ The following are the minimum software and hardware requirements to run Charmed 
 
 ### Juju
 
-The charm supports several Juju releases from [2.9 LTS](https://documentation.ubuntu.com/juju/3.6/reference/juju/juju-roadmap-and-releases/#juju-2-9) onwards. The table below shows which minor versions of each major Juju release are supported by Charmed PostgreSQL.
+The charm supports several Juju releases, starting from [2.9 LTS](https://documentation.ubuntu.com/juju/3.6/reference/juju/juju-roadmap-and-releases/#juju-2-9). The table below shows which minor versions of each major Juju release are supported by Charmed PostgreSQL.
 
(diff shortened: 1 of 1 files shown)


docs/reference/statuses.md
```diff
@@ -13,13 +13,13 @@ The charm follows [standard Juju applications statuses](https://juju.is/docs/olm
...
-| **blocked** | failed to install snap packages | There are issues with the network connection and/or the Snap Store | Check your internet connection and https://status.snapcraft.io/. Remove the application and when everything is ok, deploy the charm again |
-| **blocked** | failed to patch snap seccomp profile | The charm failed to patch one issue that happens when pgBackRest restores a backup (this blocked status should be removed when https://github.com/pgbackrest/pgbackrest/releases/tag/release%2F2.46 is added to the snap) | Remove the unit and add it back again |
+| **blocked** | failed to install snap packages | There are issues with the network connection and/or the Snap Store | Check your internet connection and https://status.snapcraft.io/. Remove the application and when everything is OK, deploy the charm again |
+| **blocked** | failed to patch snap {spellexception}`seccomp` profile | The charm failed to patch one issue that happens when pgBackRest restores a backup (this blocked status should be removed when https://github.com/pgbackrest/pgbackrest/releases/tag/release%2F2.46 is added to the snap) | Remove the unit and add it back again |
```
dragomirp, line 17: I think this status is no longer happening (we are no longer patching the profile).

dragomirp (approved):

docs/reference/statuses.md
taurus-forever: It still happens for old (initial) charm revision. Long-term we should add a column to indicate the affected revisions.

PR #1002, opened 2025-06-16
[MISC] Hba tweaks 16

dragomirp:
## Checklist


lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -35,7 +35,7 @@
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 55
+LIBPATCH = 52
```
dragomirp, line 38: Setting to stale version, so we don't release this on the v0 track.

src/backups.py
dragomirp, line 1: Linting changes.

PR #1025, opened 2025-07-03
[MISC] Set Pgbackrest spool dir to archive mount

dragomirp:
Set spool path in Pgbackrest config to use the archive volume when restoring

## Checklist

### DIFF
--- a/src/backups.py
+++ b/src/backups.py
@@ -31,6 +31,7 @@ from constants import (
     BACKUP_TYPE_OVERRIDES,
     BACKUP_USER,
     PGBACKREST_LOGROTATE_FILE,
+    POSTGRESQL_ARCHIVE_PATH,
     WORKLOAD_OS_GROUP,
     WORKLOAD_OS_USER,
 )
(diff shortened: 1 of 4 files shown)


tests/unit/test_backups.py
```diff
@@ -1813,7 +1813,8 @@ def test_render_pgbackrest_conf_file(harness, tls_ca_chain_filename):
             storage_path=harness.charm._storage_path,
             user="backup",
             retention_full=30,
-            process_max=max(cpu_count() - 2, 1),
+            process_max=2,
```
taurus-forever, line 1816: why hardcoded 2?

taurus-forever (approved):

tests/unit/test_backups.py
dragomirp: We should be getting the number of available cores from `get_available_resources()` in case of fractional limits and the output of it is patched above.

src/backups.py
```diff
@@ -1184,6 +1185,7 @@ def _render_pgbackrest_conf_file(self) -> bool:
         # Open the template pgbackrest.conf file.
         with open("templates/pgbackrest.conf.j2") as file:
             template = Template(file.read())
+        cpu_count, _ = self.charm.get_available_resources()
```
dragomirp, line 1188: Get CPU from the helper to manage fractional limits.

delgod:
Please excuse me, I was wrong, we need put spool dir to the same volume as WAL dir, but not inside WAL dir.

PR #1034, opened 2025-07-22
[DPE-7594] Remove the trigger and observe changes in pg_database

dragomirp:
* Remove the trigger updating pg_hba
* Port pg_database observer from VM

Pgbouncer WIP PR accommodating the changes: https://github.com/canonical/pgbouncer-k8s-operator/pull/602 

## Checklist


src/authorisation_rules_observer.py
```diff
@@ -74,6 +76,20 @@ def start_authorisation_rules_observer(self):
...
+                path
+                / ".."
+                / "venv"
+                / "lib"
+                / f"python{version_info.major}.{version_info.minor}"
+                / "site-packages"
+            )
+            if path.stem == "lib":
+                new_env["PYTHONPATH"] = f"{venv_path.resolve()}:{new_env['PYTHONPATH']}"
+                break
```
dragomirp, line 92: Inject the location of the venv packages when spawning the observer, so that we can use charm deps there (psycopg2). Another option is to try to activate the venv instead.

PR #1038, opened 2025-07-29
[MISC] Sync 14 to 16

dragomirp:
Sync 14 to 16:
* Remove the pg_hba trigger

## Checklist


lib/charms/postgresql_k8s/v0/postgresql.py
dragomirp, line 1: Not bumping patch, since it is to be replaced by v1.

dragomirp: We need to port predefined roles to k8s to switch to v1.

marceloneppel (approved):
Thank you so much!

Issue #1011, opened 2025-06-20
Charm intermittently gets blocked if not in active state when a relation joins


natalian98:
When 14/stable is deployed, at times postgresql-k8s fails to initialize database relation when the integration is added before the charm gets active.

## Steps to reproduce

```
juju deploy openfga-k8s --trust --channel latest/stable  <-- this is just an example, the same happens with other charms
juju deploy postgresql-k8s --trust --channel 14/stable
juju integrate postgresql-k8s openfga-k8s
```

## Expected behaviour

Postgresql gets active and provides the database relation.

## Actual behaviour

Postgres gets stuck in blocked status with message "Failed to initialize database relation". It does not happen always, but 4/5 times I was able to reproduce it.
Is there possibly some race condition when a db relation joins "too early"? If I wait for postgres to get active before adding the relation, the issue does not occur.

## Versions

Operating system: Ubuntu 22.04.5 LTS

Juju CLI: 3.6.7-genericlinux-amd64

Juju agent: 3.6.7

Charm revision: 495, but it is true also for 14/edge and 16/edge

microk8s: MicroK8s v1.32.4 revision 8155

## Log output

Juju debug log: 

```

$ juju debug-log --replay --include postgresql-k8s/0
unit-postgresql-k8s-0: 16:15:47 INFO juju.cmd running containerAgent [3.6.7 15a36bf2de0411532d4d26d07928be1f6bc0319c gc go1.24.2]
unit-postgresql-k8s-0: 16:15:47 INFO juju.cmd.containeragent.unit start "unit"
unit-postgresql-k8s-0: 16:15:47 INFO juju.worker.upgradesteps upgrade steps for 3.6.7 have already been run.
unit-postgresql-k8s-0: 16:15:47 INFO juju.worker.probehttpserver starting http server on 127.0.0.1:65301
unit-postgresql-k8s-0: 16:15:47 INFO juju.api connection established to "wss://controller-service.controller-my-controller.svc.cluster.local:17070/model/924bc132-b2a1-4e18-8dae-a2d4bba96e7b/api"
unit-postgresql-k8s-0: 16:15:47 INFO juju.worker.apicaller [924bc1] "unit-postgresql-k8s-0" successfully connected to "wss://controller-service.controller-my-controller.svc.cluster.local:17070"
unit-postgresql-k8s-0: 16:15:47 INFO juju.worker.migrationminion migration migration phase is now: NONE
[...]

Issue #927, opened 2025-04-15
Postgresql-k8s charm fails to integrate without `wait_for_idle` before integration


swetha1654:
## Description
I was facing some issues with my integration tests CI workflow the hockeypuck charm. The [workflow](https://github.com/canonical/hockeypuck-k8s-operator/actions/runs/14353802468/job/40385101069?pr=54) kept failing since the postgresql-k8s charm was failing to deploy with this error `Failed to initialize database relation`. I tried retriggering the failed workflow multiple times, but it didn't seem like an intermittent issue. I tried running the integration tests locally and it worked fine without any issues. Seems to be a problem only when i run it on GitHub CI. I was able to fix this issue by adding a [wait_for_idle](https://github.com/canonical/hockeypuck-k8s-operator/pull/49) before integrating it with the hockeypuck charm. The wait_for_idle is usually after the integration. It used to work fine all this while, i'm not sure why it stopped working now. 

## Steps to reproduce

Raise a PR in hockeypuck to remove the [fix](https://github.com/canonical/hockeypuck-k8s-operator/blob/10269d0c34288175a652cd243009dce7a24877c4/tests/integration/conftest.py#L164) and observe the integration tests.

## Expected behavior
The workflow must run successfully after removing the `wait_for_idle` before integration.

## Actual behavior

Workflow fails with error `Failed to initialize database relation`.

## Versions

Operating system: 

Juju CLI: 

Juju agent: 

Charm revision: 

microk8s: 

## Log output

Juju debug log: 
```
INFO     pytest_operator.plugin:plugin.py:943 Juju error logs:

unit-postgresql-k8s-0: 10:12:36 ERROR unit.postgresql-k8s/0.juju-log failed to change plugins: 
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-postgresql-k8s-0/charm/lib/charms/postgresql_k8s/v0/postgresql.py", line 338, in enable_disable_extensions
    with self._connect_to_database() as connection, connection.cursor() as cursor:
  File "/var/lib/juju/agents/unit-postgresql-k8s-0/charm/lib/charms/tempo_coordinator_k8s/v0/charm_tracing.py", line 1065, in wrapped_function
    return callable(*args, **kwargs)  # type: ignore
  File "/var/lib/juju/agents/unit-postgresql-k8s-0/charm/lib/charms/postgresql_k8s/v0/postgresql.py", line 153, in _connect_to_database
    connection = psycopg2.connect(
[...]

PR #1054, opened 2025-08-11
[DPE-7603][DPE-6990] Fix check in relation initialisation

marceloneppel:
## Issue
The handler for the `database_requested` event creates a new user through the primary, but checks if the current unit is has started PostgreSQL. The current unit may be the leader but not the primary, which may lead to errors when trying to access the database.

There is also the case where, for some reason, the K8s PG primary service stops redirecting traffic to the unit for a short period. This makes the charm understand that the cluster has initialised completely and can create users (but it may not due to that weird situation). This still needs further investigation.

## Solution
Use the right condition (primary endpoint being ready) in the check.

Fixes https://github.com/canonical/postgresql-k8s-operator/issues/1011
Fixes https://github.com/canonical/postgresql-k8s-operator/issues/927

## Checklist


dragomirp (approved):

taurus-forever (approved):
Wow, this is a nice catch!

PR #1060, opened 2025-08-12
[DPE-7379] custom  entity name

dragomirp:
Postgresql side of custom username implementation based on DA-176.

## Checklist


src/charm.py
```diff
@@ -2379,8 +2380,9 @@ def relations_user_databases_map(self) -> dict:
                 logger.debug(f"User {user} has no databases to connect to")
 
         # Copy relations users directly instead of waiting for them to be created
+        custom_username_mapping = self.postgresql_client_relation.get_username_mapping()
```
dragomirp, line 2383: Can't get the custom username from the followers here, so using a cache field in the app secret.

src/relations/postgresql_provider.py
```diff
@@ -71,6 +84,27 @@ def _sanitize_extra_roles(extra_roles: str | None) -> list[str]:
         extra_roles_list = [role for role in extra_roles_list if role not in ACCESS_GROUPS]
         return extra_roles_list
 
+    def get_username_mapping(self) -> dict[str, str]:
+        """Get a mapping of custom usernames by a relation ID."""
+        if username_mapping := self.charm.get_secret(APP_SCOPE, USERNAME_MAPPING_LABEL):
+            return json.loads(username_mapping)
+        return {}
+
+    def update_username_mapping(self, relation_id: int, username: str | None) -> None:
```
dragomirp, line 93: Add or delete a mapping between rel ID and custom username. Default generated usernames are ignored, since we can infer them based on rel ID as before.

src/relations/postgresql_provider.py
```diff
@@ -84,6 +118,22 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
...
+        password = None
+        try:
+            if requested_entities := event.requested_entity_secret_content:
+                for key, val in requested_entities.items():
+                    user = key
+                    password = val
+                    break
+                if user in SYSTEM_USERS:
+                    self.charm.unit.status = BlockedStatus(FORBIDDEN_USER_MSG)
+                    return
```
dragomirp, line 131: Block if a system user has been requested.

taurus-forever (approved):
Nice!
P.S. should relation user prefix `relation_id_` go to constant eventually?

src/relations/postgresql_provider.py
delgod: Should we block if the same username is requested by two relations?

delgod (changes requested):

src/relations/postgresql_provider.py
dragomirp: Updated the check to for existing users as well.

src/relations/postgresql_provider.py
```diff
@@ -286,12 +340,32 @@ def update_tls_flag(self, tls: str) -> None:
...
+                self.charm.unit.status = BlockedStatus(INVALID_EXTRA_USER_ROLE_BLOCKING_MESSAGE)
+                return
+            for relation in self.charm.model.relations.get(self.relation_name, []):
+                try:
+                    if secret_uri := self.database_provides.fetch_relation_field(
+                        relation.id, "requested-entity-secret"
+                    ):
+                        content = self.framework.model.get_secret(id=secret_uri).get_content()
+                        for key in content:
+                            if key in SYSTEM_USERS:
```
delgod, line 359: I think if the user is requested in two relations, this function is going to clean up the Blocked status which is set in `_on_database_requested` event.

dragomirp: Added the check for existing users when unblocking as well.

delgod (approved):

PR #1109, opened 2025-09-27
PgBackrest Dashboard and alert rules

Deezzir:
## Issue

Right now, it is almost impossible to track the status of backups. There is no alert to tell an operator that a backup failed. The only possible solution is to look for logs that mark a backup as failed.

Related issues:

- https://github.com/canonical/postgresql-operator/issues/272
- https://github.com/canonical/postgresql-operator/issues/356

## Solution

The changes enable the `pgbackrest_exporter` as a service, add [a Grafana Dashboard](https://grafana.com/grafana/dashboards/17709-pgbackrest-exporter-dashboard/) for the exported metrics, and introduce a new set of alert rules to track failed backups.

In addition, the PR adds a CI job that validates the alert rules and unit tests them.

## Checklist
- [X] I have added or updated any relevant documentation.
- [X] I have cleaned any remaining cloud resources from my accounts.


src/charm.py
```diff
@@ -1827,6 +1830,18 @@ def _generate_metrics_service(self) -> dict:
             },
         }
 
+    def _generate_pgbackrest_metrics_service(self) -> dict:
+        """Generate the pgbackrest metrics service definition."""
+        return {
+            "override": "replace",
+            "summary": "pgbackrest metrics exporter",
+            "command": "/usr/bin/pgbackrest_exporter",
```
Deezzir, line 1838: We cannot use `/start-pgbackrest-exporter.sh` at the moment because the current script is not prepared to be run in a non-snapped environment, [reference](https://github.com/canonical/charmed-postgresql-snap/blob/14/edge/snap/local/start-pgbackrest-exporter.sh).

marceloneppel (approved):
Thanks a lot for this great work, @Deezzir! I left one minor suggestion.

src/charm.py
Deezzir: https://github.com/canonical/charmed-postgresql-snap/issues/169

Deezzir:
@taurus-forever, looking at logs, I see that `postgresql-test-app` fails, not the main charm

marceloneppel:
@Deezzir, adding a like the following one right before https://github.com/Deezzir/postgresql-k8s-operator/blob/db8f13a4c07b513554faebac4a896fe29160011f/tests/integration/ha_tests/test_restart.py#L41 should make the failing tests pass:
```python
await ops_test.model.relate(DATABASE_APP_NAME, f"{APPLICATION_NAME}:database")
```

PR #1115, opened 2025-09-30
Refresh v3 implementation

marceloneppel:
## Issue
The charm is not yet supporting refresh through refresh v3.

## Solution
Added refresh v3 implementation based on the [developer docs](https://canonical-charm-refresh.readthedocs-hosted.com/latest/) and also based on the implementations from [PostgreSQL VM](https://github.com/canonical/postgresql-operator/pull/866) and [Kafka K8s](https://github.com/canonical/kafka-k8s-operator/pull/200).

Use `set_unit_status` to set the unit status and properly set statuses during refresh.

The old refresh/upgrade logic/files were removed.

Correctly reload pgBackRest TLS server configuration with SIGHUP to avoid interrupting the data replication between the primary and the replica (more details in the comments on this PR).

After this PR is approved, https://github.com/canonical/postgresql-k8s-operator/pull/1115/commits/2a806f7bbd432e8ee654b8ba4e47aefa9dc40a61 must be reverted, then the PR should me merged, to properly generate the tags in the repo (this is the same as stated at https://github.com/canonical/postgresql-operator/pull/866#issuecomment-2854799705).

## Checklist


src/backups.py
```diff
@@ -1303,7 +1306,18 @@ def start_stop_pgbackrest_service(self) -> bool:
...
-        self.container.restart(self.charm.pgbackrest_server_service)
+        services = self.container.pebble.get_services(names=[self.charm.pgbackrest_server_service])
+        if len(services) == 0:
+            return False
+
+        if services[0].current == ServiceStatus.ACTIVE:
+            logger.debug("Sending SIGHUP to pgBackRest TLS server to reload configuration")
+            self.container.pebble.send_signal(
+                signal.SIGHUP, services=[self.charm.pgbackrest_server_service]
+            )
```
marceloneppel: Always restarting the service was causing a SIGTERM to be sent to the pgBackRest TLS  server in the middle of the process where the replica gets the latest data from the primary to create a backup. This started to happen because of some more deferred events happening. A SIGHUP is enough to update the TLS server when the certificates change.

tests/integration/helpers.py
```diff
@@ -820,6 +821,22 @@ async def switchover(
             assert standbys == len(ops_test.model.applications[app_name].units) - 1
 
 
+async def switchover_to_unit_zero(ops_test: OpsTest) -> None:
```
marceloneppel: Used to reduce the number of switchovers to the minimum (as we expect since the previous refresh/upgrade implementation).

taurus-forever (approved):

dragomirp (approved):

src/charm.py
```diff
@@ -251,18 +252,13 @@ def __init__(self, *args):
         self.framework.observe(self.on.promote_to_primary_action, self._on_promote_to_primary)
         self.framework.observe(self.on.get_primary_action, self._on_get_primary)
         self.framework.observe(self.on.update_status, self._on_update_status)
+        self.framework.observe(self.on.collect_unit_status, self._reconcile_refresh_status)
```
carlcsaposs-canonical: (just realized I should've added this in an inline comment in the vm implementation https://github.com/canonical/postgresql-operator/pull/866)

fyi this shouldn't cause any issues as-is but if collect_unit_status is observed more than once, ops will prioritize statuses incorrectly. more info: https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/status/#implementation

src/refresh.py
```diff
@@ -0,0 +1,69 @@
...
+        """Implement pre-refresh checks before any unit refreshed."""
+        self.run_pre_refresh_checks_after_1_unit_refreshed()
+
+        # If the first unit is not the primary we ask the user to switchover
+        # the primary to it.
+        primary_unit_name = self._charm._patroni.get_primary(unit_name_pattern=True)
+        unit_zero_name = f"{self._charm.app.name}/0"
+        if primary_unit_name != unit_zero_name:
+            raise charm_refresh.PrecheckFailed(
+                f"Switch primary to {unit_zero_name} to avoid multiple switchovers during refresh."
```
carlcsaposs-canonical: if a pre-refresh failure message is greater than 64 characters, it will get truncated in juju status
recommend <=64 characters, if possible
https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/pre-refresh-checks/#implement-check

carlcsaposs-canonical (commented):
mostly looks good! couple questions about HA

* could the workload be started when workload_allowed_to_start is False? e.g. in `_on_peer_relation_changed` -> `self.update_config()` or in deferred config-changed/pebble-ready during teardown
  https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/workload-allowed-to-start/
* question: for now, will scaling up be declared as not supported in the user docs? if the intention is to support it, PeerRelationNotReady may need to be handled differently: https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/peer-relation-not-ready/
* remember to update the promote workflow (either in this PR or I can follow up in another PR) before the first promotion https://github.com/canonical/postgresql-k8s-operator/pull/1146#discussion_r2513642677

src/refresh.py
```diff
@@ -0,0 +1,69 @@
...
+            )
```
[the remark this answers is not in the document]
marceloneppel: AFAIK, refreshing the primary before non-primary units might lead to incompatibility in data replication format in some cases.

It might be interesting to run the check before each unit is refreshed, except before the last unit (unit 0). Is `run_pre_refresh_checks_before_any_units_refreshed` the place for doing that?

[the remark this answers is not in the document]
carlcsaposs-canonical: also, in the VM charm we currently do the switchover automatically instead of telling the user to do it

not sure which approach is better, but should we use the same approach on both charms? (e.g. switch VM to instruct user or switch K8s to do it automatically)

carlcsaposs-canonical (approved):
looks great!

comments are minor & non-blocking

src/backups.py
```diff
@@ -820,6 +821,8 @@ def _on_create_backup_action(self, event) -> None:  # noqa: C901
                 "pgbackrest",
                 f"--stanza={self.stanza_name}",
                 "--log-level-console=debug",
+                "--log-level-file=debug",
```
taurus-forever, line 824: https://warthogs.atlassian.net/browse/DPE-9935

marceloneppel: https://github.com/canonical/postgresql-k8s-operator/pull/1485

PR #1151, opened 2025-10-29
[DPE-8550] Check juju version before removing revision and remove discourse charmers test

dragomirp:
* Check juju version before removing a secret revision
* Remove legacy admin relation test

## Checklist

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -19,20 +19,24 @@ jobs:
         uses: actions/checkout@v5
       - name: Set up environment
         run: |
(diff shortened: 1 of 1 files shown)


marceloneppel (approved):

taurus-forever (approved):

tests/integration/test_db_admin.py
dragomirp: The discourse charmers charm has not been updated since 2021, is superseded by a newer charm and fails to deploy locally and on CI. I don't think there's another db-admin charm to use, so removing the test.

PR #1166, opened 2025-11-27
[DPE-8986] Check for database not running only in pre-checks and fix test

marceloneppel:
## Issue
When a refresh is performed, and the refresh is incompatible (like on a PR), the first refreshed unit won’t start the database due to the refresh checks. This makes the upgrade tests get stuck in the following state (from https://github.com/canonical/postgresql-k8s-operator/actions/runs/19736056202/job/56548782683#step:6:2541):
```sh
postgresql-k8s/2 [idle] blocked: Rollback with `juju refresh`. Pre-refresh check failed: PostgreSQL is not running on 1+ units
```

## Solution
In that case, the check for the database not running shouldn’t run, because the database will be started only after the refresh checks pass. So that check was moved to `run_pre_refresh_checks_before_any_units_refreshed`, which won't run after the first unit is refreshed.

This PR will be updated if https://github.com/canonical/postgresql-k8s-operator/pull/1149 is merged first.

## Checklist

### DIFF
--- a/src/refresh.py
+++ b/src/refresh.py
@@ -67,7 +67,7 @@ class PostgreSQLRefresh(CharmSpecificKubernetes):
             member_name = f"{self._charm.app.name}-{unit_number}"
             if unit_number != highest_unit_number and member_name not in running_members:
                 raise charm_refresh.PrecheckFailed(
-                    f"Unit {member_name} is not online. All units except the highest unit must be online before refresh."
+                    f"PostgreSQL is not running on unit {unit_number}"
                 )
 
         # Switch primary to last unit to refresh (lowest unit number).


tests/integration/ha_tests/test_upgrade_from_stable.py
dragomirp: Merged the new jubilant tests, we may not need `run-pre-refresh-checks` override in [here](https://github.com/canonical/postgresql-k8s-operator/blob/16/edge/tests/integration/high_availability/high_availability_helpers_new.py#L292)

carlcsaposs-canonical (approved):

PR #1167, opened 2025-11-28
juju tf/1.0

shipperizer:
- **docs: add tf readme from tf-docs command**


taurus-forever (approved):
Finally we had a chance to check this PR, it has been lost without default reviewers for some reason. Assigned the complete list here.

LGTM, THANK YOU!!!!
one question inside.

PR #1192, opened 2025-12-19
Enhance update_status to monitor health checks and collect diagnostic logs

marceloneppel:
## Solution
Previously, the update_status hook only checked if the PostgreSQL pebble service was active. This could miss situations where the service appeared running but the health check was failing. This change enhances the restart logic to also monitor Pebble health check status and collect diagnostic information when issues are detected.

Changes:
- Monitor Pebble health check status in addition to service status
- Trigger restart when health check is DOWN even if service is ACTIVE
- Add _get_postgresql_startup_logs() to collect diagnostic information from Pebble logs and PostgreSQL logs before and after restart attempts
- Add comprehensive unit tests for health check monitoring and error handling

This improves observability and helps diagnose PostgreSQL restart issues by capturing relevant logs at the time of failure.

## Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,9 +15,9 @@ on:
       - '**.md'
       - .github/renovate.json5
       - 'docs/**'
-  schedule:
-    - cron: '53 0 * * *'  # Daily at 00:53 UTC
-  # Triggered on push to branch "main" by .github/workflows/release.yaml
+  # Triggered on schedule by .github/workflows/scheduled_ci_*.yaml on default branch
+  workflow_dispatch:
+  # Triggered on push by .github/workflows/release.yaml
   workflow_call:
     outputs:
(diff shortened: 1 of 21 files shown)


marceloneppel:
> Nice PR and interesting [findings](https://github.com/canonical/postgresql-k8s-operator/actions/runs/20370209233/job/58539052302?pr=1192#step:6:5885) right out of the box: `ERROR cannot start service while killing`

Hopefuly, this PR is the solution for the issue that we saw when upgrading PG to 14.20. I believe the new PG is working a little bit different and Pebble from Juju 2.9 is not correctly detecting that Patroni is stopped (it seems that Pebble is seeing PG still running and considering the Patroni service - PG process parent - still active because of that).

Regarding the finding you commented, I believe that error may be a side effect of the fix added by this PR. It will need more investigation.

PR #1235, opened 2026-02-04
[MISC] Port pgbackrest exporter

Deezzir:
Backport of the following commit - cb19f4c

## Checklist


marceloneppel (commented):
Thanks for the amazing work, @Deezzir!

Something I noticed was that the dashboard uses [`pgbackrest_exporter_build_info`](https://github.com/canonical/postgresql-k8s-operator/blob/a76decb/src/grafana_dashboards/postgresql-pgbackrest-metrics.json#L6888), but this metric was [introduced in v0.20.0](https://github.com/woblerr/pgbackrest_exporter/releases/tag/v0.20.0) when `pgbackrest_exporter_info` was renamed.

The snap currently pins [`prometheus-pgbackrest-exporter=0.17.0`](https://github.com/canonical/charmed-postgresql-snap/blob/636222df1a8bd383ad7593897f52cebc9f013944/snap/snapcraft.yaml#L226), which only exposes [`pgbackrest_exporter_info`](https://github.com/woblerr/pgbackrest_exporter/blob/v0.17.0/README.md#L72).

This most probably will cause the dashboard panel(s) using this metric to show no data.

Deezzir:
@marceloneppel Do you think it is reasonable to try to bump the exporter version? And you're right, I think currently we're using the dashboard `rev6` which works from exporter version `0.22`. It is also the case for the charm on `14/stable` track. It is a mistake on my part. [Reference](https://grafana.com/grafana/dashboards/17709-pgbackrest-exporter-dashboard/)

marceloneppel (approved):
LGTM! I''m in sync with @taurus-forever's above comment.

dragomirp (approved):

PR #1266, opened 2026-02-16
[DPE-7891] test: re-enable discourse integration test

marceloneppel:
## Issue

The Discourse integration test (`test_discourse`) was disabled with `pytest.skip("Second migration doesn't complete")`. The skip was carried over from the `main` branch, where a `PrematureDataAccessError` in `update_tls_flag` caused the test to fail. The `16/edge` branch is not affected by this bug — its `update_connection_info` method already guards against uninitialized relations with `if not database or not password: continue`.

## Solution

Re-enable the Discourse integration test by removing the `pytest.skip`. The test passes since the code path that caused the failure on `main` does not exist on this branch.

## Checklist

### DIFF
--- a/tests/integration/new_relations/test_new_relations_2.py
+++ b/tests/integration/new_relations/test_new_relations_2.py
@@ -64,7 +64,6 @@ async def test_database_deploy_clientapps(ops_test: OpsTest, charm):
 
 @markers.amd64_only  # discourse-k8s charm not available for arm64
 async def test_discourse(ops_test: OpsTest):
-    pytest.skip("Second migration doesn't complete")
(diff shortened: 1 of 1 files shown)


carlcsaposs-canonical (approved):

taurus-forever (approved):

dragomirp (approved):

marceloneppel:
> @marceloneppel how expensive is to guard main in the same/similar way?

@taurus-forever, I believe it's not needed to guard main because it already has a guard implemented. We just forgot to re-enable the test on that branch, too. I commented about that in https://github.com/canonical/postgresql-k8s-operator/pull/1265.

PR #1284, opened 2026-02-18
[DPE-9448] Don't override upgrade status

dragomirp:
## Issue
Unit status during upgrade gets overwritten, hiding from the user that upgrade did not complete

## Solution
Add a check for upgrade idleness in the helper method

## Checklist

### DIFF
--- a/src/relations/db.py
+++ b/src/relations/db.py
@@ -13,14 +13,17 @@ from charms.postgresql_k8s.v0.postgresql import (
     PostgreSQLDeleteUserError,
     PostgreSQLGetPostgreSQLVersionError,
 )
-from ops.charm import (
+from ops import (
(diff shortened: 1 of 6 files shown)


src/relations/db.py
```diff
@@ -13,14 +13,17 @@
     PostgreSQLDeleteUserError,
     PostgreSQLGetPostgreSQLVersionError,
 )
-from ops.charm import (
+from ops import (
```
dragomirp: Import from ops directly, not strictly necessary.

carlcsaposs-canonical (approved):

marceloneppel (approved):

taurus-forever (approved):
LGTM. I expected bigger changes here. Tnx!

PR #1330, opened 2026-03-09
[DPE-9568] fix(tests): avoid race in test_no_prefix_dbs and retry CLI errors in test_symlinks

marceloneppel:
## Issue
Two integration test flakes identified from [Allure report run 5430](https://canonical.github.io/postgresql-k8s-operator/16_edge/5430/ ).

### Issue 1: test_no_prefix_dbs race condition

`test_no_prefix_dbs` removes two relations (di1 and di2) then calls `wait(all_agents_idle)` once. Juju processes the resulting hooks (relation-departed, relation-broken) sequentially, and the unit briefly appears idle between dispatches. The poll returns in that gap before all hooks complete, so `prefix-databases` is still present in the relation databag.

### Issue 2: test_symlinks CLIError

`test_build_and_deploy` in `test_symlinks.py` fails intermittently with `jubilant.CLIError: connection is shut down` during `juju.wait()`. Other tests in the codebase already handle this with `retry_if_cli_error`.

The remaining two issues still need investigation.

## Solution
Wait for idle after each relation removal instead of removing both, then waiting once — Juju briefly appears idle between sequential hook dispatches.

Wrap `juju.wait()` with `retry_if_cli_error` — the same pattern already used in other integration tests.

## Checklist

### DIFF
--- a/tests/integration/test_symlinks.py
+++ b/tests/integration/test_symlinks.py
@@ -14,6 +14,7 @@ from .helpers import (
     DATABASE_APP_NAME,
     METADATA,
 )
+from .jubilant_helpers import retry_if_cli_error
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


dragomirp (approved):

taurus-forever (approved):
Great job done!

PR #1342, opened 2026-03-10
[DPE-9580] fix: set up pgdata symlinks in reconcile, add CLI retry, and re-enable upgrade tests

marceloneppel:
## Issue

1. `test_extra_user_roles` in `test_invalid_extra_user_roles.py` fails intermittently with `jubilant.CLIError: connection is shut down` during `juju.wait()`, as seen in [Allure report run 5434](https://canonical.github.io/postgresql-k8s-operator/16_edge/5434/).

2. Upgrade tests fail because `reconcile()` restarts Patroni (via `_update_pebble_layers(replan=True)`) without first creating the symlink `/var/lib/postgresql/16/main → /var/lib/pg/data/16/main`. On non-first units during upgrade, `workload_allowed_to_start` is already `True` when the pod restarts, so `reconcile()` fires immediately during `upgrade-charm` — before `pebble-ready` gets a chance to create the symlinks. This causes `pg_basebackup` to fail with `Permission denied` when trying to create `/var/lib/postgresql/16/` (owned by root).

3. Upgrade tests were disabled (excluded via spread `systems`). The initial deploy channel should point to `16/beta` now that it's available.

## Solution

1. Wrap `juju.wait()` with `retry_if_cli_error` — the same pattern already used in other integration tests. We'll most probably need to use a different approach, maybe create some kind of implicit wrapper in the juju object to always retry instead of wrapping each juju object call.

2. Split `_create_pgdata` into two methods: `_create_pgdata` (which conditionally clears stale WAL/temp directories on replicas, then delegates) and `_ensure_pgdata_dirs_and_symlinks` (which idempotently creates storage directories, symlinks, and fixes permissions). Call only `_ensure_pgdata_dirs_and_symlinks` from `reconcile()` to ensure symlinks exist before Patroni starts, without running the destructive stale-directory cleanup during upgrades. The cleanup is safe to skip during upgrades because `PG_VERSION` persists on PV-backed storage. See [DPE-9581](https://warthogs.atlassian.net/browse/DPE-9581) for a follow-up to replace the delete with a move.

3. Re-enable all four upgrade spread tests and set the initial deploy channel to `16/beta`.

## Checklist

[DPE-9581]: https://warthogs.atlassian.net/browse/DPE-9581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -328,7 +328,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
     def reconcile(self):
         """Reconcile the unit state on refresh."""
         self.set_unit_status(MaintenanceStatus("starting services"))
-        self._create_pgdata(self._container)
+        self._ensure_pgdata_dirs_and_symlinks(self._container)
         self._update_pebble_layers(replan=True)
 
         if not self._patroni.member_started:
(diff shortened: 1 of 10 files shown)


taurus-forever (commented):
LGTM, nice to see all tests passed within 1 re-try!
Can you decrease risks for reconcile call?

taurus-forever (dismissed):

carlcsaposs-canonical (approved):

taurus-forever (approved):

PR #1359, opened 2026-03-17
[DPE-9630] fix: verify replication HBA before starting Patroni on replicas

marceloneppel:
## Issue

When a replica's endpoint is added to the cluster members list, the leader writes a new `patroni.yml` and reloads the primary's Patroni to update `pg_hba` with the replica's replication entry. However, the replica's deferred `pebble-ready` handler can re-fire before that reload completes, causing `pg_basebackup` to be rejected. Patroni then calls `remove_data_directory()`, which unlinks the pgdata symlink, and the subsequent `pg_basebackup` creates a real directory on ephemeral storage instead of writing through the symlink to the PVC.

This means `readlink -f /var/lib/postgresql/16/main` returns `/var/lib/postgresql/16/main` (real directory) instead of `/var/lib/pg/data/16/main` (symlink to PVC), causing `test_symlinks.py` to fail and — more critically — PostgreSQL data to be written to ephemeral storage that is lost on pod restart.

Additionally, `test_data_replication` and `test_standby_promotion` in the async replication tests lacked `all_active` waits before querying units, causing intermittent `Connection refused` errors when PostgreSQL was still starting.

### Add replication HBA check to `_replica_can_start()`

A third gate is added to `_replica_can_start()`: before allowing Patroni to start, the charm attempts a physical replication connection (using `psycopg2.extras.PhysicalReplicationConnection`) to the primary. If the connection succeeds, the primary's `pg_hba` reload has taken effect and `pg_basebackup` will not be rejected. This eliminates the race at source.

The fix was validated by running `test_symlinks.py` twice on a fresh 3-unit deployment (7/7 passed both times, including `test_pgdata_symlinks[2]` which previously failed), all unit tests passing (`tox run -e unit`), and all linting passing (`tox run -e lint`).

### Add `all_active` waits to async replication tests

`test_data_replication` and `test_standby_promotion` now wait for `all_active` before querying units, preventing `Connection refused` errors during PostgreSQL startup.

## Checklist


dragomirp (dismissed):

dragomirp (dismissed):

carlcsaposs-canonical (commented):
> When Patroni reports `member_started=True` but the health state is `"starting"` (PostgreSQL not yet accepting connections), `_set_active_status` now sets `WaitingStatus("waiting for PostgreSQL to start")` instead of `ActiveStatus`. This prevents premature readiness signaling. The WaitingStatus is:
> * **Skipped during refresh** (so the refresh library doesn't consider the unit unhealthy)

Does this mean the next unit could refresh before the current unit is accepting connections?

If so, is the reason for this because the unit that hasn't started isn't guaranteed to get another Juju event after it has started? (If so, would it make sense to hold execution in the event until PostgreSQL is started?)

marceloneppel:
> Does this mean the next unit could refresh before the current unit is accepting connections?
> 
> If so, is the reason for this because the unit that hasn't started isn't guaranteed to get another Juju event after it has started? (If so, would it make sense to hold execution in the event until PostgreSQL is started?)

The actual gate preventing the next unit from refreshing is `next_unit_allowed_to_refresh` in the relation databag, set at the end of `reconcile()` only after `is_replication_healthy()` succeeds (which itself has a 60-second internal retry checking Patroni's `/leader` and `/replica?lag=16kB` endpoints for all members). A unit in "starting" state fails that check, so `reconcile()` cannot set the flag until the unit is genuinely streaming. The refresh library then reads this flag to decide whether to lower the StatefulSet partition and allow the next unit to update — it never inspects unit status messages.

The readiness signaling I mentioned is for test checks.

carlcsaposs-canonical (dismissed):

taurus-forever (changes requested):

carlcsaposs-canonical (approved):

taurus-forever (approved):

PR #1444, opened 2026-04-13
[DPE-9845] fix: use real pgdata path for Patroni to fix reinit and move logs to persistent storage

marceloneppel:
## Issue

When Patroni performs `reinit` on a replica via `patronictl reinit`, its `remove_data_directory()` internally runs `rm -rf` on its configured `data_dir`, which was previously set to the Debian-style symlink path `/var/lib/postgresql/16/main`. Because that path is a symlink pointing to `/var/lib/pg/data/16/main`, `rm -rf` only removes the symlink itself rather than the real directory it points to. As a result, the actual data, logs, and temp subdirectories under `/var/lib/pg/data/16/main` are left intact, causing Patroni reinit to fail because the directory is not wiped as expected.

Additionally, PostgreSQL, Patroni, and pgBackRest logs were written to `/var/log/postgresql` and `/var/log/pgbackrest` on ephemeral storage, meaning logs were lost across container restarts.

## Solution

**Fix Patroni reinit:** Configure Patroni's `data_dir` to point directly to the real pgdata path (`/var/lib/pg/data/16/main`, i.e. `_actual_pgdata_path`) instead of the Debian symlink path. With this change, `remove_data_directory()` operates directly on the actual directory and correctly removes all data, logs, and temp subdirectories. Reinit then succeeds as expected.

The Debian-style compatibility symlink is also moved one level up: `/var/lib/postgresql/16 -> /var/lib/pg/data/16` (instead of `/var/lib/postgresql/16/main -> /var/lib/pg/data/16/main`). This preserves `/var/lib/postgresql/16/main` as a valid alias to pgdata for any tools that rely on the traditional Debian path layout, while ensuring that traversal through the symlink also correctly resolves the real directory.

The now-unused `self.pgdata_path` attribute and `POSTGRESQL_DATA_PATH` constant are removed; all consumers now use `self._actual_pgdata_path` directly.

**Move logs to persistent storage:** Relocate log directories to the persistent `logs` storage volume:
- PostgreSQL: `/var/lib/pg/logs/16/main/pg_logs`
- Patroni: `/var/lib/pg/logs/16/main/patroni_logs`
- pgBackRest: `/var/lib/pg/logs/16/main/pgbackrest_logs`

The legacy `/var/log/postgresql` path is preserved as a symlink pointing to the PostgreSQL logs directory. Upgrade handling is included: if `/var/log/postgresql` exists as an empty directory from a previous deployment it is replaced with the symlink; if it is a non-empty directory or an unexpected file type, the unit surfaces a clear error and blocks until manually resolved.

**Testing:** A new integration test (`test_patroni_reinit_after_pg_control_deletion`) validates the reinit fix end-to-end: it simulates a corrupted replica by deleting `pg_control`, triggers `patronictl reinit`, and asserts the replica recovers to a healthy streaming state with data integrity intact. `test_symlinks.py` is renamed to `test_filesystem_layout.py` to reflect its broader scope.

## Checklist


src/charm.py
```diff
@@ -1215,28 +1215,26 @@ def _ensure_pgdata_dirs_and_symlinks(self, container: Container):
...
         # Note: This symlink is on ephemeral storage and may not persist across container restarts.
         # It gets recreated on each pebble-ready event.
-        container.make_dir(
-            "/var/lib/postgresql/16",
-            user=WORKLOAD_OS_USER,
-            group=WORKLOAD_OS_GROUP,
-            make_parents=True,
-        )
+        # /var/lib/postgresql is created by the postgresql Debian package and exists in the
+        # container image, so no make_dir is needed before creating the symlink.
```
taurus-forever, line 1299: mkdir for available folder has no harm :-)

taurus-forever (dismissed):
Try #2 🤞 
Thank you!

dragomirp (dismissed):

taurus-forever:
@marceloneppel is it ready for review/testing?

marceloneppel:
> @marceloneppel is it ready for review/testing?

@taurus-forever, not yet. I'm working on making the creation of the logs symlink safer (by not unconditionally removing the directory if it exists in the same path).

marceloneppel:
Closing and re-opening the PR to try to run all CI tests.

dragomirp (approved):

src/charm.py
```diff
@@ -1197,6 +1255,21 @@ def _ensure_pgdata_dirs_and_symlinks(self, container: Container):
...
+        for path in [POSTGRESQL_LOGS_PATH, PATRONI_LOGS_PATH, PGBACKREST_LOGS_PATH]:
+            if not container.exists(path):
+                container.make_dir(
+                    path,
+                    permissions=0o755,
+                    user=WORKLOAD_OS_USER,
+                    group=WORKLOAD_OS_GROUP,
+                    make_parents=True,
+                )
+            container.exec(["chmod", "755", path]).wait()
```
taurus-forever, line 1267: The last 5 (for others) might be revised in the future.

taurus-forever (approved):
Let's test it practically in 16/edge.

PR #1463, opened 2026-04-14
[DPE-9705] Limit max_timelines_history to 50

taurus-forever:
## Issue

The 'history' part of patroni config enpoint grows without limits.

## Solution

Limit the history to 50 records. 640Kb will be enough for everyone (c)

Fixes https://github.com/canonical/postgresql-k8s-operator/issues/1388


templates/patroni.yml.j2
```diff
@@ -5,6 +5,7 @@ bootstrap:
     synchronous_mode: true
     synchronous_mode_strict: false
     synchronous_node_count: {{ synchronous_node_count }}
+    max_timelines_history: 50
```
dragomirp, line 8: We need to add it to bulk_update calls and refresh code, so old deployments will pick up the new value.

taurus-forever: Any references to follow? :cry:

taurus-forever: @dragomirp it is 16/edge K8s, we do not have previous stable releases. Are changes still necessary?

dragomirp: Should be fine then.

tests/unit/test_patroni.py
```diff
@@ -233,6 +233,7 @@ def test_render_patroni_yml_file(harness, patroni):
             rewind_password=patroni._rewind_password,
             maximum_lag_on_failover=1048576,
             synchronous_node_count=0,
+            max_timelines_history=50,
```
dragomirp: Is this necessary? The value should be hardcoded in the template itself.

dragomirp (dismissed):

tests/unit/test_patroni.py
taurus-forever: Removed in [01fd44f](https://github.com/canonical/postgresql-k8s-operator/pull/1463/commits/01fd44f69448e23aa3458d715a3a3b396201fd3d)

marceloneppel (approved):

dragomirp (approved):

PR #1551, opened 2026-05-28
fix(tests): correct the location of the Harness cleanup

tonyandrewmeyer:
## Issue

The `test_get_extensions` test calls `cleanup()` on Harness before using it. We're fixing a bug in Harness that leaves a dangling sqlite3 reference (https://github.com/canonical/operator/pull/2507), and that fix breaks this test because of this.

## Solution

Call `cleanup` when finished with the harness.

## Checklist
- [ ] ~I have added or updated any relevant documentation.~
- [ ] ~I have cleaned any remaining cloud resources from my accounts.~

### DIFF
--- a/tests/unit/test_db.py
+++ b/tests/unit/test_db.py
@@ -223,12 +223,12 @@ def test_get_extensions(harness):
         default: false
         type: boolean"""
     harness = Harness(PostgresqlOperatorCharm, config=config)
-    harness.cleanup()
     harness.begin()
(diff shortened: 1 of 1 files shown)


carlcsaposs-canonical (approved):
thank you!

taurus-forever (approved):
Nice catch!

marceloneppel (approved):
Thank you, @tonyandrewmeyer!

dragomirp (approved):

dragomirp (approved):

PR #1581, opened 2026-06-09
fix: make _create_pgdata idempotent on re-emitted pebble-ready

taurus-forever:
## Issue

Nightly CI/CD test failed:
https://github.com/canonical/postgresql-k8s-operator/actions/runs/27175388869/job/80223717923#logs

```
unit-postgresql-k8s-0: 00:29:23 ERROR unit.postgresql-k8s/0.juju-log database-peers:0: Uncaught exception while in charm code:
  File "/var/lib/juju/agents/unit-postgresql-k8s-0/charm/src/charm.py", line 2810, in <module>
    main(PostgresqlOperatorCharm, use_juju_for_storage=True)
...
  File "/var/lib/juju/agents/unit-postgresql-k8s-0/charm/venv/lib/python3.10/site-packages/ops/pebble.py", line 2673, in _raise_on_path_error
    raise PathError(error['kind'], error['message'])
ops.pebble.PathError: generic-file-error - mkdir /var/lib/postgresql/data/pgdata: file exists
```

## Solution

A deferred _on_postgresql_pebble_ready event is re-emitted by
the ops framework on every subsequent hook.
On re-emit, _create_pgdata runs again; container.exists() can report the
pgdata directory as missing even when present (restrictively-mounted cloud volumes),
so make_dir() raises "PathError: ... file exists" and crashes the hook.

Make _create_pgdata idempotent by tolerateting the redundant make_dir() by
adding make_parents=True (like happens for 16/edge).

Added unit tests to prevent regression.

Assisted-by: Claude:claude-4.8-opus

## Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -933,21 +933,13 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
     def _create_pgdata(self, container: Container):
         """Create the PostgreSQL data directory."""
         if not container.exists(self.pgdata_path):
-            try:
-                container.make_dir(
-                    self.pgdata_path,
-                    permissions=0o750,
-                    user=WORKLOAD_OS_USER,
-                    group=WORKLOAD_OS_GROUP,
-                )
-            except PathError as e:
-                # This handler can run more than once, e.g. a deferred pebble-ready event is
-                # re-emitted on later hooks. container.exists() may also report the pgdata
-                # directory as missing even when it is present (restrictively-mounted cloud
-                # volumes), so a redundant make_dir() must stay idempotent: ignore "file exists".
-                if "file exists" not in str(e):
-                    raise
-                logger.info("pgdata directory already exists, skipping creation")
+            container.make_dir(
+                self.pgdata_path,
+                permissions=0o750,
+                user=WORKLOAD_OS_USER,
+                group=WORKLOAD_OS_GROUP,
+                make_parents=True,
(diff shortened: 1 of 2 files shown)


taurus-forever:
For the history, 16/edge is not affected due to `make_parents=True` there:
```
  if not container.exists(self._actual_pgdata_path):
      container.make_dir(
          self._actual_pgdata_path,
          permissions=0o700,
          user=WORKLOAD_OS_USER,
          group=WORKLOAD_OS_GROUP,
          make_parents=True,        # <-- this is the key difference
      )
```

Why that closes the bug: make_parents=True gives Pebble mkdir -p / os.MkdirAll semantics. When the target already exists as a directory, it returns success instead of raising PathError: … file exists. So the exact failure mode from main — a re-emitted deferred pebble-ready re-running creation while container.exists() wrongly reports the dir missing — can't crash here: even if the exists() guard is bypassed, the redundant make_dir is a no-op rather than an EEXIST.

carlcsaposs-canonical (approved):
could be worth using charmlibs-pathops in the future

which has pathlib.Path semantics, including an `exist_ok=True` argument that does not also create parents

https://documentation.ubuntu.com/charmlibs/reference/charmlibs/pathops/#pathops.ContainerPath.mkdir

marceloneppel (approved):

dragomirp (approved):

PR #1618, opened 2026-06-26
[DPE-10614] Migrate TLS to the single-kernel library (K8s)

marceloneppel:
## Issue

The K8s charm owns its TLS subsystem (`src/relations/tls.py`, `push_tls_files_to_workload`) duplicated from the VM charm and the single-kernel library's now-equivalent `events.tls.TLS` + `TLSManager`.

## Solution

Migrates the K8s charm's TLS onto the library's operator-cert handler (`events.tls.TLS`) + `TLSManager`. The events handler owns the two `TLSCertificatesRequiresV4` requirers and is constructor-injected into `TLSManager`, whose live-fetch getters read cert/key from them (no operator-cert peer secrets — live-fetch, rollback-clean).

- Bump `postgresql-charms-single-kernel` from 16.3.2 to 16.3.4 to consume the lib TLS handler + manager; redirect the operator-cert getters to `tls_manager`.
- Reload bridge `_reload_tls_after_push`: after the lib stores+pushes certs, refresh the K8s container trust store and charm-local CA bundle (`_sync_tls_trust_store_and_bundle`) and reload; also fires on `relation_broken` so detaching the TLS operator re-renders Patroni with TLS off.
- On the pebble-ready bootstrap path and in `_fix_pod`, the lib's `generate_internal_peer_cert` does not push (unlike the old charm method), so the charm now explicitly calls `tls_manager.push_tls_files()` + `_sync_tls_trust_store_and_bundle()` + `update_config()` to render Patroni with TLS on (`_fix_pod` guarded on the internal-ca secret).
- Remove `src/relations/tls.py` and `push_tls_files_to_workload` (now lib-owned).
- Rewrote `tests/unit/test_tls.py` to cover the lib-backed wiring (reload bridge, trust-store sync, ordering, defer guards) instead of the removed `relations.tls.TLS` module.

The bridge and its four observers are transitional — they stay charm-side only until the `update_config`/Patroni-config subsystem ports to the library, at which point push+reload become a single lib-owned atomic step and these charm methods are removed.

Also, applied the `_re_emit_async_relation_changed_event` `StopIteration` guard with a regression test.

## Checklist


src/charm.py
```diff
@@ -296,7 +294,31 @@ def __init__(self, *args):
...
+        )
+        self.framework.observe(
+            self.tls.peer_certificate.on.certificate_available, self._reload_tls_after_push
+        )
+        self.framework.observe(
+            self.on[TLS_CLIENT_RELATION].relation_broken, self._reload_tls_after_push
+        )
+        self.framework.observe(
+            self.on[TLS_PEER_RELATION].relation_broken, self._reload_tls_after_push
+        )
```
dragomirp, line 321: IMHO it would be cleaner to pass the `_reload_tls_after_push` function as a parameter to the `TLSManager` constructor, instead of relying on ops handler order being consistent across versions and architectures.

marceloneppel: Agreed that the reload shouldn't depend on observer order. The order itself is actually guaranteed (documented in `Framework.observe`, and the wiring tests here assert it), but while checking that I found a worse problem: defer is per-observer. If the lib handler defers its push (e.g. a Pebble write fails during a cert renewal) while this handler completes (the files-on-disk guard passes on the old files), then the retried push writes the new cert and nothing reloads. The old charm did push+reload in one handler, so a defer retried both.

I'd put the hook one level up from `TLSManager` though. `push_tls_files()` has no event, so a callback there can't defer when `update_config` fails, and pebble-ready/`_fix_pod` call the manager directly and already do their own reload, so it would fire twice there. A `tls_files_pushed` custom event on the lib's TLS handler (like `refresh_tls_certificates_event`) may avoid both, and the two guards this handler duplicates can then go away.

That needs a lib change and a release, so I'd keep this PR as-is and do it as a follow-up: lib PR, then rewire both charm PRs. Or I can take it in this one if you prefer.

taurus-forever (approved):
LGTM, good catch Drago!

dragomirp (approved):

PR #1629, opened 2026-07-06
[DPE-9978] fix(charm): apply minimal ownerReferences SSA patch on app removal

marceloneppel:
## Issue

On stop, the charm reparented Patroni/Juju Endpoints+Services onto the StatefulSet (pod0's owner) by server-side-applying the WHOLE object listed moments earlier. Patroni mutates its `patroni-postgresql-k8s-config` Service (its DCS config object) continuously, so the snapshot carried a stale `metadata.resourceVersion`. Server-Side Apply (SSA) treats a body `resourceVersion` as an optimistic-concurrency precondition, so the apply 409'd with "the object has been modified; please apply your changes to the latest version", the reparent silently failed (the apply is in a `except ApiError: logger.exception` block), and the resource leaked — k8s garbage-collects objects whose `ownerReferences` points at the StatefulSet when it is deleted on app removal, so a resource that never got reparented is never GC'd.

## Solution

Apply a targeted patch carrying only the object's name/namespace (identity) + `ownerReferences` instead. SSA skips the `resourceVersion` precondition (absent from the body) and leaves Patroni-owned fields untouched (SSA only touches fields present in the patch body), so `ownerReferences` reparenting succeeds and GC proceeds on StatefulSet deletion. `force=True` is retained so the charm can take ownership of `ownerReferences` even if another field manager owns it. Omitting `resourceVersion` makes the precondition that caused this 409 structurally impossible to trigger, not merely less likely.

## Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1960,21 +1960,10 @@ class PostgresqlOperatorCharm(TypedCharmBase[K8SCharmConfig]):
                 or resource.metadata.ownerReferences == pod0.metadata.ownerReferences
             ):
                 continue
-            # Reparent the resource onto the StatefulSet (pod0's owner) so it
-            # is garbage-collected when the application is removed.
-            #
-            # Apply only ownerReferences (plus identity) via a targeted
-            # server-side-apply patch, NOT the whole listed object. Patroni
-            # mutates its leader-election Service (e.g. config annotations)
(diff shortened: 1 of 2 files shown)


dragomirp (approved):

taurus-forever (approved):
Thank you!

PR #1657, opened 2026-07-14
[DPE-10731] fix(terraform): allow juju provider 1.x and 2.x

marceloneppel:
## Issue

The terraform module pinned the juju provider to `~> 1.0` (`>= 1.0.0, < 2.0.0`).

There is a conflict in the **v2 cap**: a consumer requiring juju provider 2.x (current v2.1.1; supports Juju 3.x/4.x) hits the `< 2.0.0` bound and fails `terraform init`. The v1→v2 upgrade is non-breaking for this module's resource surface (a single `juju_application`).

Separately, `terraform/**` was in `ci.yaml`'s `paths-ignore`, so a broken constraint would land with no checks at all.

## Solution

**Widen the constraint** — `terraform/versions.tf`: `~> 1.0` → `>= 1.0, < 3.0`, admitting v1.x and v2.x while capping an unvetted v3. Regenerate the `terraform/README.md` tables and fix its Usage examples (they referenced a non-existent `juju_model_name` var and a wrong `storage` shape).

**Add a terraform test lane** (mirrors `canonical/iam-bundle-integration`'s static checks and `canonical/mysql-plans`' deploy test):
- `.github/workflows/terraform.yaml` — runs on `terraform/**` (the inverse of the path-ignore); installs terraform via snap and runs the `lint-terraform` (fmt/init/validate) and `terraform` (composition) tox envs, both locally runnable.
- `tests/terraform/test_compositions.py` — 4 cases sourcing the real module beside a sibling constraint (`~> 1.0`, `>= 1.1.1`, `>= 2.0`, unsatisfiable `>= 99.0`), asserting `init` resolves the expected provider major or fails cleanly. Regression guard for the widened constraint.
- `tests/integration/terraform/test_terraform_deploy.py` + spread task — jubilant test applying the module (published `16/stable` charm) into the `testing` model, driving storage/config/constraints, waiting for active/idle, and asserting the `application_name` output. A task-local spread variant (`tfprov_v1`) reruns it pinned to `~> 1.0`, so the module is deployed under **both** provider lines, asserting the resolved major each time.
- `TF_BINARY` override and `terraform/.gitignore`.

## Checklist


tests/spread/test_terraform_deploy.py/task.yaml
```diff
@@ -0,0 +1,14 @@
+summary: terraform/test_terraform_deploy.py
+environment:
+  TEST_MODULE: terraform/test_terraform_deploy.py
+  # tfprov_v1: task-local variant (only this task gains it; others stay juju36). CONCIERGE_JUJU_CHANNEL
+  # is repeated because this variant doesn't inherit the global juju36-scoped value.
+  CONCIERGE_JUJU_CHANNEL/tfprov_v1: 3.6/stable
+  TF_PROVIDER_CONSTRAINT/tfprov_v1: "~> 1.0"
```
carlcsaposs-canonical, line 7: is the idea to introduce other variants in the future?

if not, would it be simpler to remove task-specific variants?

tests/terraform/test_compositions.py
```diff
@@ -0,0 +1,97 @@
...
+import pytest
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+TERRAFORM_MODULE = REPO_ROOT / "terraform"
+TF_BINARY = os.getenv("TF_BINARY") or "terraform"
+TF_TIMEOUT = 5 * 60
+
+pytestmark = pytest.mark.skipif(
+    shutil.which(TF_BINARY) is None, reason=f"{TF_BINARY} not found on PATH"
+)
```
carlcsaposs-canonical, line 28: might be worth failing instead of skipping if terraform is not correctly installed, to avoid false sense of security from workflow run passing

carlcsaposs-canonical (approved):

dragomirp (approved):

tests/spread/test_terraform_deploy.py/task.yaml
marceloneppel: You're right — the variant is more machinery than this needs. Dropped it in https://github.com/canonical/postgresql-k8s-operator/pull/1673: the deploy test is now parametrized on `[v2, v1]` in one spread job, with a `try/finally` that `terraform destroy`s between params (`restore-each` only runs between jobs, not between params, so without it the legs would collide on the `postgresql-k8s` app in the shared model). The parametrized test IDs (`[v2]`/`[v1]`) also fix the Allure visibility — the variant's `:tfprov_v1` results were dropped by the `endsWith ':juju36'` upload gate because both legs shared one pytest test ID. One job instead of two, no task-local variant env. No future variants were planned; this was a one-off for the v1 line the widened constraint admits.

tests/terraform/test_compositions.py
marceloneppel: Done in https://github.com/canonical/postgresql-k8s-operator/pull/1673 — switched the static composition test's guard from `pytest.mark.skipif` to a module-level `pytest.fail`, so a missing terraform binary fails the lane instead of leaving it green with all cases skipped. The integration deploy test keeps its skip — local-skip is legitimate for an integration test run outside CI, whereas the static lane is a dedicated terraform CI job where the binary is mandatory.

PR #1759, opened 2026-09-15
[MISC] Assert refresh version (14/edge)

dragomirp:
CI test to assert that workload version matches the refresh definition version.

Also remove a duplicate function.


src/relations/db.py
```diff
@@ -130,14 +130,6 @@ def _on_relation_changed(self, event: RelationChangedEvent) -> None:
...
         return any(r in ALL_LEGACY_RELATIONS for r in self.charm.client_relations)
 
-    def _check_multiple_endpoints(self) -> bool:
-        """Checks if there are relations with other endpoints."""
-        is_exist = self._check_exist_current_relation()
-        for relation in self.charm.client_relations:
-            if relation.name not in ALL_LEGACY_RELATIONS and is_exist:
-                return True
-        return False
-
```
marceloneppel, line 140: Is this removal intentional?

dragomirp: Yes, it's redefined on line 392, but I forgot to write a comment.

marceloneppel: Good catch. Thanks!

marceloneppel (approved):

carlcsaposs-canonical (approved):

PR #1760, opened 2026-09-16
[DPE-11059] Check for relation app before trying to remove

dragomirp:
Workaround for missing app name.


src/relations/tls_transfer.py
```diff
@@ -59,7 +59,7 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
     def _on_certificate_removed(self, event: CertificateRemovedEvent) -> None:
         """Disable TLS when TLS certificate is removed."""
         relation = self.charm.model.get_relation(TLS_TRANSFER_RELATION, event.relation_id)
-        if relation is None:
+        if relation is None or relation.app is None:
```
dragomirp, line 62: Should fix the immediate issue, but we should most probably not cache the CA in a secret at all.

carlcsaposs-canonical (approved):

marceloneppel (approved):

